PR #61, opened 2017-01-30
Release 0.2.1

maxhutch:
Lots of code cleaning and robustness improvements.

### DIFF
--- a/src/main/scala/io/citrine/lolo/bags/Bagger.scala
+++ b/src/main/scala/io/citrine/lolo/bags/Bagger.scala
@@ -108,12 +108,17 @@ class BaggedTrainingResult(
   lazy val NibT = Nib.transpose
   lazy val model = new BaggedModel(models, Nib, useJackknife, biasModel)
   lazy val rep = trainingData.head._2
-  lazy val predictedVsActual = trainingData.zip(NibT).map { case ((f, l), nb) =>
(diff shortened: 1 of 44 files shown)


src/main/scala/io/citrine/lolo/bags/Bagger.scala
```diff
@@ -108,12 +108,17 @@ class BaggedTrainingResult(
...
+  lazy val predictedVsActual = trainingData.zip(NibT).flatMap { case ((f, l), nb) =>
+    val oob = models.zip(nb).filter(_._2 == 0)
+    if (oob.nonEmpty) {
+      val predicted = rep match {
+        case x: Double => oob.map(_._1.transform(Seq(f)).getExpected().head.asInstanceOf[Double]).sum / oob.size
+        case x: Any => oob.map(_._1.transform(Seq(f)).getExpected().head).groupBy(identity).maxBy(_._2.size)._1
+      }
+      Seq((f, predicted, l))
+    } else {
+      Seq()
```
maxhutch: @eantono this is the substantial block

maxhutch:
There shouldn't be an issue with 1 tree: it would be equivalent to a single train-test split with 37% in the test split, so we'd still get a RMSE.

eantono:
Oh yeah I see.

Issue #69, opened 2017-03-08
Feature importances should factor in linear model weights for linear leaves


eantono:
Once standardization is in place, we should think about how to factor linear model weights into feature importances.

maxhutch:
Agreed.  The tree importances are based on changes in impurity across a split.  In that vein, we could take the remaining impurity at the leaf and distribute it across the features in the leaf model weighted by their internal importance.  This would be _easy_ to implement.

Thoughts?  CC: @jling11 @sparadiso

sparadiso:
That seems reasonable.

PR #70, opened 2017-03-08
Bubbling leaf importance up through tree

maxhutch:
This is a prototype in response to #69 

@eantono do you have a good case for tests?


eantono:
@maxhutch can you use the thermoelectrics demo with the CV splits that don't group on temperature?

PR #71, opened 2017-03-13
Release 0.2.3

maxhutch:
This release adds standardization and leaf importance propagation.

### DIFF
--- a/src/main/scala/io/citrine/lolo/transformers/Standardizer.scala
+++ b/src/main/scala/io/citrine/lolo/transformers/Standardizer.scala
@@ -1,0 +1,221 @@
+package io.citrine.lolo.transformers
+
+import io.citrine.lolo.{Learner, Model, PredictionResult, TrainingResult}
+
+/**
+  * Standardize the training data to zero mean and unit variance before feeding it into another learner
+  *
+  * This is particularly helpful for regularized methods, like ridge regression, where
+  * the relative scale of the features and labels is important.
+  *
+  * Created by maxhutch on 2/19/17.
+  */
+class Standardizer(baseLearner: Learner) extends Learner() {
+  override var hypers: Map[String, Any] = Map()
+
+  override def setHypers(moreHypers: Map[String, Any]): this.type = {
+    baseLearner.setHypers(moreHypers)
+    super.setHypers(moreHypers)
+  }
+
+  /**
+    * Create affine transformations for continuous features and labels; pass data through to learner
+    *
+    * @param trainingData to train on
+    * @param weights      for the training rows, if applicable
+    * @return training result containing a model
+    */
+  override def train(trainingData: Seq[(Vector[Any], Any)], weights: Option[Seq[Double]]): StandardizerTrainingResult = {
+    val rep = trainingData.head._1
+    val inputTrans = Standardizer.getMultiStandardization(trainingData.map(_._1))
+    val outputTrans = trainingData.head._2 match {
+      case x: Double => Some(Standardizer.getStandardization(trainingData.map(_._2.asInstanceOf[Double])))
+      case x: Any => None
+    }
+
+    val (inputs, labels) = trainingData.unzip
+    val standardTrainingData = Standardizer.applyStandardization(inputs, inputTrans).zip(Standardizer.applyStandardization(labels, outputTrans))
+    val baseTrainingResult = baseLearner.train(standardTrainingData)
+
+    new StandardizerTrainingResult(baseTrainingResult, Seq(outputTrans) ++ inputTrans, hypers)
+  }
+}
+
+/**
+  * Training result wrapping the base learner's training result next to the transformations
+  *
+  * @param baseTrainingResult
+  * @param trans
+  * @param hypers
+  */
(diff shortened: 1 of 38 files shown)


maxhutch:
Rebuild me!

src/main/scala/io/citrine/lolo/transformers/Standardizer.scala
```diff
@@ -0,0 +1,221 @@
...
+      case Some(x) if trans.head.isDefined => Some(x.map(_.asInstanceOf[Double] * rescale))
+      case x: Any => x
+    }
+  }
+
+
+  /**
+    * Get the gradient or sensitivity of each prediction
+    *
+    * This is un-stanardized by rescaling by the label's variance divided by the feature transformation's variance
```
jling11: typo

src/main/scala/io/citrine/lolo/transformers/Standardizer.scala
```diff
@@ -0,0 +1,221 @@
...
+  val intercept = trans.head.map(_._1).getOrElse(1.0)
+}
+
+/**
+  * Utilities to compute and apply standarizations
+  */
+object Standardizer {
+
+  /**
+    * The standardizations are just shifts and rescale.  The shift is by the mean and the re-scale is by the variance
```
jling11: It might be nice to think about extensibility for different types of scalers (e.g. min-max scalers.)

src/main/scala/io/citrine/lolo/trees/regression/RegressionTrainingLeaf.scala
```diff
@@ -0,0 +1,41 @@
...
+  * Created by maxhutch on 3/8/17.
+  */
+class RegressionTrainingLeaf(
+                              trainingData: Seq[(Vector[AnyVal], Double, Double)],
+                              leafLearner: Learner,
+                              depth: Int
+                            ) extends TrainingNode(trainingData, depth) {
+
+  /**
+    * Average the training data
```
jling11: I wouldn't mind more descriptive code comments in this class.

src/main/scala/io/citrine/lolo/trees/regression/RegressionTrainingLeaf.scala
```diff
@@ -0,0 +1,41 @@
...
+    new ModelLeaf(leafLearner.train(trainingData).getModel().asInstanceOf[Model[PredictionResult[Double]]], depth)
+  }
+
+  def getFeatureImportance(): scala.collection.mutable.ArraySeq[Double] = {
+    importance match {
+      case Some(x) =>
+        val expectations: (Double, Double) = trainingData.map{ case (v, l, w) =>
+          (l * w, l * l * w)
+        }.reduce((u: (Double, Double), v: (Double, Double)) => (u._1 + v._1, u._2 + v._2))
+        val impurity = expectations._2 / trainingData.size - Math.pow(expectations._1 / trainingData.size, 2.0)
```
jling11: This is not relevant to this particular code review, but why isn't the impurity stored at each node?  Why do we have to recalculate it here, when it was already calculated when the node was split?

src/test/scala/io/citrine/lolo/trees/regression/RegressionTreeTest.scala
```diff
@@ -148,6 +147,25 @@ class RegressionTreeTest {
     val oos = new ObjectOutputStream(new FileOutputStream(tmpFile))
     oos.writeObject(DT)
   }
+
+  /**
+    * Test a really short tree to make sure the linear model feature importance gets carried through
+    */
+  def testShortTreeWithLinearLeaf(): Unit = {
+     val trainingData = TestUtils.generateTrainingData(1024, 12, noise = 0.1, function = Friedman.friedmanSilverman)
+
```
jling11: It might also be good to do another test with a tree with a single leaf node, so that it's deterministic (can you turn off bagging for your trees?).  This test is good, but since it's not deterministic, it might good to also have a deterministic one.

src/main/scala/io/citrine/lolo/transformers/Standardizer.scala
maxhutch: I plan on revisiting these patterns when we have another motivated transformer.

src/main/scala/io/citrine/lolo/trees/regression/RegressionTrainingLeaf.scala
maxhutch: We don't actually calculate the impurity in the splitters.  This is a performance optimization: the splitter considers a very large number of splits, so the least it does for each one the better.

The importance isn't always used, so why compute it unless it is?  This is more of a memory optimization (feature importance can be larger than the split) than a run-time one.

We could expose other importance measures, but I'm not sure that using impurity is on a critical performance path (see memory comment above).  Although we don't compute the impurity for each split, we do compute the change in impurity, so that's already available.

maxhutch:
@jling11 thanks for the comments!  I made some changes.  Thoughts?

PR #72, opened 2017-03-24
Hotfix - standardizer was incorrect

maxhutch:
The standardizer wasn't passing the weights through, which was really important to the functioning of bagged models.


maxhutch:
CC: @sparadiso this was a serious issue but a simple fix.

PR #74, opened 2017-04-20
Interruptible bagger training

maxhutch:
Bagger training can take a long time.  If its in a webservice, that can cause problems.  We should be able to interrupt it if we need to.

This PR establishes two patterns for making routines interruptible:
 * Interruptible parallel collections via the `InterruptibleExecutionContext`
 * Consistent thread status checking with `Async.canStop`

The unit test can only make sure the `train` method is interrupted _somewhere_.  I welcome suggestions for checking how quickly it stops running.

CC: @sparadiso

### DIFF
--- a/src/main/scala/io/citrine/lolo/bags/Bagger.scala
+++ b/src/main/scala/io/citrine/lolo/bags/Bagger.scala
@@ -4,8 +4,10 @@ import breeze.linalg.{DenseMatrix, DenseVector, min, norm}
 import breeze.numerics.abs
 import breeze.stats.distributions.Poisson
 import io.citrine.lolo.stats.metrics.ClassificationMetrics
+import io.citrine.lolo.util.InterruptibleExecutionContext
 import io.citrine.lolo.{Learner, Model, PredictionResult, TrainingResult}
 
+import scala.collection.parallel.ExecutionContextTaskSupport
 import scala.collection.parallel.immutable.ParSeq
 
 /**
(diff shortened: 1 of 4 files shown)


maxhutch:
Rebuild me!

src/main/scala/io/citrine/lolo/bags/Bagger.scala
```diff
@@ -61,10 +63,13 @@ class Bagger(
     }
 
     /* Learn the actual models */
-    val models = (0 until actualBags).par.map { i =>
+    val parIterator = (0 until actualBags).par
+    parIterator.tasksupport = new ExecutionContextTaskSupport(InterruptibleExecutionContext())
```
sparadiso: I like this a lot!

src/main/scala/io/citrine/lolo/bags/Bagger.scala
```diff
@@ -324,15 +333,21 @@ class BaggedResult(
 
     /* These operations are pulled out of the loop and extra-verbose for performance */
     val JMat = NibJ.t * predMat
+    Async.canStop()
     val JMat2 = JMat :* JMat * ((Nib.size - 1.0) / Nib.size)
+    Async.canStop()
     val IJMat = NibIJ.t * predMat
+    Async.canStop()
```
sparadiso: No reason not to sprinkle these around liberally, but as long as we can bound time to interrupt below ~1-2secs, it's fine.

src/test/scala/io/citrine/lolo/bags/BaggerTest.scala
```diff
@@ -100,6 +102,39 @@ class BaggerTest {
...
+        override def call() = {
+          println("Starting to run the future")
+          val res = baggedLearner.train(trainingData)
+          assert(false, "Training was not terminated")
+          res
+        }
+      }
+    )
+    Thread.sleep(1000)
+    assert(fut.cancel(true), "Failed to cancel future")
```
sparadiso: Can you add a timer to ensure that the time it takes from line 126 to 136 is <2secs or so?

sparadiso:
yesssssss. RE your last point - the easy thing to do is to time how long it takes to get past `future.get()`. If you wanted to go deeper, you could wrap the whole thing in `tryToInterruptAfter(delay: Int)`, which sleeps for different amounts of time to measure the variance in time-to-interrupt.

PR #75, opened 2017-04-21
Refactor the training to discard each model's meta-data asap

maxhutch:
This should reduce the peak memory usage during training.

CC: @eantono @sparadiso

### DIFF
--- a/src/main/scala/io/citrine/lolo/bags/Bagger.scala
+++ b/src/main/scala/io/citrine/lolo/bags/Bagger.scala
@@ -32,6 +32,14 @@ class Bagger(
 
   override var hypers: Map[String, Any] = Map()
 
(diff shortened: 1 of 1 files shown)


src/main/scala/io/citrine/lolo/bags/Bagger.scala
```diff
@@ -98,7 +105,7 @@ class Bagger(
       val biasModel = biasLearner.get.train(biasTraining).getModel()
       Async.canStop()
 
-      new BaggedTrainingResult(models.map(_.getModel()), hypers, importances, Nib, trainingData, useJackknife, Some(biasModel))
+      new BaggedTrainingResult(models, hypers, averageImportance, Nib, trainingData, useJackknife, Some(biasModel))
```
sparadiso: noice

PR #81, opened 2017-04-25
Hotfix: system blas can be pickier about inversion.

maxhutch:
The pseudo-inverse is always a safe choice.


maxhutch:
Note: there was no version bump because I hadn't deployed yet.

PR #84, opened 2017-05-21
Adding tests to expose bug in linear model feature importance

maxhutch:
Categorical values were being dropped from the feature importance.


maxhutch:
Thanks for the catch @jling11

PR #85, opened 2017-06-06
Fix bug in leaf importance calculation

maxhutch:
The local variance computation, which is used to weight the importance,
has a weights-related bug that results in negative importance.

CC: @jling11

### DIFF
--- a/src/main/scala/io/citrine/lolo/trees/regression/RegressionTrainingLeaf.scala
+++ b/src/main/scala/io/citrine/lolo/trees/regression/RegressionTrainingLeaf.scala
@@ -31,10 +31,10 @@ class RegressionTrainingLeaf(
   def getFeatureImportance(): scala.collection.mutable.ArraySeq[Double] = {
     importance match {
       case Some(x) =>
-        val expectations: (Double, Double) = trainingData.map{ case (v, l, w) =>
-          (l * w, l * l * w)
-        }.reduce((u: (Double, Double), v: (Double, Double)) => (u._1 + v._1, u._2 + v._2))
-        val impurity = expectations._2 / trainingData.size - Math.pow(expectations._1 / trainingData.size, 2.0)
+        val expectations: (Double, Double, Double) = trainingData.map{ case (v, l, w) =>
+          (l * w, l * l * w, w)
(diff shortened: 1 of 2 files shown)


src/test/scala/io/citrine/lolo/trees/regression/RegressionTreeTest.scala
```diff
@@ -187,6 +207,6 @@ object RegressionTreeTest {
     new RegressionTreeTest().longerTest()
     new RegressionTreeTest().testCategorical()
     new RegressionTreeTest().testLinearLeaves()
-    new RegressionTreeTest().testShortTreeWithLinearLeaf()
```
jling11: Don't you want to add the testStumpwithLinearLeaves test ?

src/test/scala/io/citrine/lolo/trees/regression/RegressionTreeTest.scala
```diff
@@ -173,6 +175,24 @@ class RegressionTreeTest {
     /* They should have roughly the same value (linear regression has some noise in it) */
     assert(Math.abs(importances(0) - 0.30030792928576033) < 1.0e-3)
   }
+
+  /**
+    * Test with random weights
```
jling11: "Test with random weights to ensure all feature importances of stump tree with linear leaves are non-negative."

sparadiso:
LGTM - @jling11 ?

src/main/scala/io/citrine/lolo/trees/regression/RegressionTrainingLeaf.scala
```diff
@@ -31,10 +31,10 @@ class RegressionTrainingLeaf(
   def getFeatureImportance(): scala.collection.mutable.ArraySeq[Double] = {
```
jling11: I would add on to this comment with the formula used to calculate the impurity now:
sum(L**2 w)/ sum(w) - (sum(L*w))**2/w**2, where L is the regression label, W is the weight.

src/test/scala/io/citrine/lolo/trees/regression/RegressionTreeTest.scala
maxhutch: This is just a driver for human use.  The tests are run whenever they have `@Test` annotations

PR #88, opened 2017-06-08
Differences of small numbers have noise

maxhutch:
In case the impurity is a small amplitude negative number, round up to zero.


sparadiso:
LGTM - is this good to merge?

PR #90, opened 2017-06-27
Release 0.2.8

maxhutch:
This is almost exclusively automatic deployments from master with travis.


cd/before-deploy.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+if [ "$TRAVIS_BRANCH" = 'master' ] && [ "$TRAVIS_PULL_REQUEST" == 'false' ]; then
```
sinistral: What is the significance of TRAVIS_PULL_REQUEST == false?

src/test/scala/io/citrine/lolo/PerformanceTest.scala
```diff
@@ -47,8 +47,7 @@ class PerformanceTest {
     (timeTraining, timePredicting)
   }
 
-  @Test
-  @Category(Array(classOf[SlowTest]))
+  // @Category(Array(classOf[SlowTest]))
```
sinistral: Remove?

cd/before-deploy.sh
maxhutch: We only want to setup and execute the deployment on builds on the master branch.  In travis, when you open a release PR into master, it spawns 3 builds: one on the branch, one on the merged branch before the merge, and one on the `master` branch after the merge.  This just picks out only the last, post-merge build.

src/test/scala/io/citrine/lolo/PerformanceTest.scala
maxhutch: I would like to add that test back after a refactor in the future.  It currently times out on travis, so it needs some work.  That refactor is only worth doing if this travis-based deployment works, which is why I haven't invested the time in it already.

cd/before-deploy.sh
sinistral: That's a nice system.  Thanks for the explanation.

src/test/scala/io/citrine/lolo/PerformanceTest.scala
sinistral: I have often found it useful to drop a link to an issue opened to track the work alonside the commented out code.  It helps to raise awareness in sprint planning so that it isn't forgotten and gives people a place to discuss the change. 

YMMV.

cd/before-deploy.sh
maxhutch: I copy-pasted, but this sort of travis environment variable inspection is common.  Travis takes the 1 config per project approach (vs separate PR and deploy configs in jenkins), so it is sure to expose [nice stuff](https://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables).

sinistral:
I approve, but apparently as a reviewer without write access my opinion doesn't count.

PR #93, opened 2017-07-04
Moving timings to the theta library

maxhutch:
Re-enabling the performance test and re-basing it on the [theta](https://github.com/CitrineInformatics/theta) performance analysis tool.


maxhutch:
The un-run test is a disabled jenkins test.  I'll remove it later; overriding to merge now since the other two tests passed.

Issue #83, opened 2017-05-17
Allow max_depth = 0


jling11:
It should be possible to have no splits in a tree (to get just bagged guessthemean or bagged linearregression).

maxhutch:
Resolved in #94

PR #94, opened 2017-07-05
Set `maxDepth` properly and check that it is obeyed (it is not)

maxhutch:
These tests should fail.


maxhutch:
This will close #83.

PR #95, opened 2017-07-06
Adding support for Boltzmann trees

maxhutch:
This is done via a splitter that selects splits with a probability
that is computed as exp[-(\Delta Impurity) / temperature], where
the temperature is a hyperparameter that controls the balance
between robustness and precision


maxhutch:
@sparadiso and @jling11  this isn't ready to merge, but do you have thoughts on the methodology?

src/main/scala/io/citrine/lolo/trees/splits/AnnealingSplitter.scala
```diff
@@ -0,0 +1,196 @@
...
+
+import scala.collection.immutable.BitSet
+import scala.collection.mutable
+import scala.util.Random
+
+/**
+  * Find a split for a regression problem
+  *
+  * The splits are picked with a probability that is related to the reduction in variance:
+  * P(split) ~ exp[ - {reduction in variance} / ({temperature} * {total variance} ]
```
sparadiso: Missing end paren

src/main/scala/io/citrine/lolo/trees/splits/AnnealingSplitter.scala
```diff
@@ -0,0 +1,196 @@
...
+        0.0
+      }
+      val pivot = (thinData(j + 1)._1 + thinData(j)._1) / 2.0
+      (score, pivot, totalVariance)
+    }
+
+    val totalScore = possibleSplits.map(_._1).sum
+    val draw = Random.nextDouble() * totalScore
+    var cumSum: Double = 0.0
+    possibleSplits.foreach{case (score, pivot, variance) =>
```
jling11: I like this logic.  Adding a comment explaining how it works could help future generations.

src/main/scala/io/citrine/lolo/trees/splits/AnnealingSplitter.scala
```diff
@@ -0,0 +1,196 @@
...
+      } else {
+        0
+      }
+
+      val includeSet: mutable.BitSet = new mutable.BitSet() ++ orderedNames.slice(0, j + 1).map(_.toInt)
+
+      (score, includeSet, totalVariance)
+    }
+
+    val totalScore = possibleSplits.map(_._1).sum
```
jling11: Is there any case where this totalScore would be zero?

src/main/scala/io/citrine/lolo/trees/splits/AnnealingSplitter.scala
```diff
@@ -0,0 +1,196 @@
...
+    val draw = Random.nextDouble() * totalScore
+    var cumSum: Double = 0.0
+    possibleSplits.foreach{case (score, includeSet, variance) =>
+        cumSum = cumSum + score
+        if (draw < cumSum) {
+          return (new CategoricalSplit(index, includeSet), variance, totalScore)
+        }
+    }
+    throw new RuntimeException("Draw was beyond all the probabilities")
```
jling11: When would this exception get thrown?

maxhutch: It should never get thrown; its almost a comment.  I'll add a real comment too.

src/test/scala/io/citrine/lolo/AccuracyTest.scala
```diff
@@ -0,0 +1,132 @@
...
+    val baseLearner = new RegressionTreeLearner(numFeatures = nFeat / 3).setHyper("minLeafInstances", minInstances)
+    val learner = new Bagger(baseLearner, numBags = nRow * nScal)
+    val error = computeMetrics(learner)
+    assert(error > noiseLevel, s"Can't do better than noise")
+    println(s"Error in RF is ${error}")
+  }
+
+  @Test
+  def testLowTLimit(): Unit = {
+    println("Hello?")
```
andrewmillspaugh: Hello!

src/main/scala/io/citrine/lolo/trees/splits/AnnealingSplitter.scala
maxhutch: Commented

src/main/scala/io/citrine/lolo/trees/splits/AnnealingSplitter.scala
maxhutch: I changed the return value of the helper splitting methods to be options.  Now, in any case where a returned split would have zero probability, it actually returns None.  That way, we can "check for zero probability" by checking if the returned splits have any non-None (by flattening them and seeing if the result has any splits in it).

src/main/scala/io/citrine/lolo/trees/splits/AnnealingSplitter.scala
maxhutch: Added.  I'd like to DRY it out with a helper too, but I don't think that's low hanging fruit right now.

maxhutch:
This is finally ready @mrupp-citrine @sparadiso @bfolie @WardLT !

pom.xml
```diff
@@ -6,7 +6,7 @@
 
     <groupId>io.citrine</groupId>
     <artifactId>lolo</artifactId>
-    <version>2.2.5</version>
+    <version>0.2.9</version>
```
sparadiso: 2.3.0?

python/lolopy/learners.py
```diff
@@ -316,36 +316,40 @@ class RandomForestClassifier(BaseLoloClassifier, RandomForestMixin):
 class RegressionTreeLearner(BaseLoloRegressor):
     """Regression tree learner, based on the RandomTree algorithm."""
 
-    def __init__(self, num_features=-1, max_depth=30, min_leaf_instances=1, leaf_learner=None,
-                 randomize_pivot_location=False):
```
maxhutch: Note: we can still use randomize_pivot_locations in lolopy using the `RandomForestRegressor`.  This only removes control over the pivot locations when working with a single regression tree in lolopy (which I don't think is a super-high priority).  We could add this and general lolopy control over the splitters in another PR.

pom.xml
maxhutch: This is a PR into develop, so it shouldn't be changing at all.

src/test/scala/io/citrine/lolo/AccuracyTest.scala
```diff
@@ -0,0 +1,127 @@
...
+        numFeatures = nFeat,
+        splitter = BoltzmannSplitter(temperature = 1.0e-9)
+      )
+      val learner = new Bagger(baseLearner, numBags = nRow * 16)
+      // println(s"Annealing train time: ${Stopwatch.time(computeMetrics(learner))}")
+      computeMetrics(learner)
+    }
+
+    val relativeDifference = 2.0 * Math.abs(errorAnnealingTree - errorStandardTree)/(errorAnnealingTree + errorStandardTree)
+    println(relativeDifference)
```
bfolie: Remove print statement?

src/test/scala/io/citrine/lolo/AccuracyTest.scala
```diff
@@ -0,0 +1,127 @@
...
+    val pva: Seq[(Double, Double, Double)] = labels.indices.map{i =>
+      (labels(i), expected(i), sigma(i))
+    }
+    val rmse = Math.sqrt(pva.map{case (x, y, s) => Math.pow(x-y, 2)}.sum / pva.size)
+    val stdres = Math.sqrt(pva.map{case (x, y, s) => Math.pow((x-y) / s, 2)}.sum / pva.size)
+    (rmse, stdres)
+  }
+
+
+  def main(args: Array[String]): Unit = {
```
bfolie: Is this meant to be a test or is it just for diagnostics?

src/test/scala/io/citrine/lolo/AccuracyTest.scala
```diff
@@ -0,0 +1,127 @@
...
+
+/**
+  * Driver code to study the performance vs temperature
+  */
+object AccuracyTest {
+
+  val trainingDataFull: Seq[(Vector[Any], Double)] = TestUtils.binTrainingData(TestUtils.generateTrainingData(2048, 48), inputBins = Seq((2, 32)))
+    .asInstanceOf[Seq[(Vector[Any], Double)]]
+
+  def computeMetrics(nRow: Int, nFeat: Int, nFeatSub: Int, nScal: Int, minInstances: Int, temperature: Double): (Double, Double) = {
```
bfolie: Could you add a docstring stating which metrics are computed?

src/main/scala/io/citrine/lolo/trees/splits/BoltzmannSplitter.scala
```diff
@@ -0,0 +1,226 @@
...
+  * The splits are picked with a probability that is related to the reduction in variance:
+  * P(split) ~ exp[ - {reduction in variance} / ({temperature} * {total variance}) ]
+  * recalling that the "variance" here is weighted by the sample size (so its really the sum of the square difference
+  * from the mean).  This is akin to kinetic monte carlo and simulated annealing techniques.
+  *
+  * Created by maxhutch on 11/29/16.
+  *
+  * @param temperature used to control how sensitive the probability of a split is to its change in variance
+  */
+case class BoltzmannSplitter(temperature: Double) extends Splitter[Double] {
```
bfolie: Should there be a hard requirement that `temperature > 0`?

src/main/scala/io/citrine/lolo/trees/splits/BoltzmannSplitter.scala
```diff
@@ -0,0 +1,226 @@
...
+import scala.collection.mutable
+import scala.util.Random
+
+/**
+  * Find a split for a regression problem
+  *
+  * The splits are picked with a probability that is related to the reduction in variance:
+  * P(split) ~ exp[ - {reduction in variance} / ({temperature} * {total variance}) ]
+  * recalling that the "variance" here is weighted by the sample size (so its really the sum of the square difference
+  * from the mean).  This is akin to kinetic monte carlo and simulated annealing techniques.
```
bfolie: I'm confused as to exactly which values are weighted and which are normalized.
Can I interpret `temperature` consistently independent of scale and sample size? My default interpretation is "if `temperature = 0.02`, then any split that reduces the relative variance by about 2% or more is equally likely to be picked." Is that correct? And if I train a model with a given temperature and then train a different model where I've scaled all the labels by a factor of 10 or used 10 times as much training data (drawn from the same underlying distribution), can I use the same temperature and expect the same results?

src/main/scala/io/citrine/lolo/trees/splits/BoltzmannSplitter.scala
```diff
@@ -0,0 +1,226 @@
+package io.citrine.lolo.trees.splits
+
+import io.citrine.lolo.trees.impurity.VarianceCalculator
+import io.citrine.lolo.trees.splits.BoltzmannSplitter.SplitterResult
+
+import scala.collection.mutable
+import scala.util.Random
+
+/**
+  * Find a split for a regression problem
```
bfolie: What are the expected benefits of Boltzmann trees? Why is introducing more randomness into which split is picked expected to improve model performance? Would temperature be a hyperparameter that automatically gets optimized?

src/main/scala/io/citrine/lolo/trees/splits/BoltzmannSplitter.scala
```diff
@@ -0,0 +1,226 @@
...
+    val draw = Random.nextDouble() * totalProbability
+    // could be a scanLeft + find, but this is more readable
+    var cumSum: Double = 0.0
+    possibleSplits.foreach { case SplitterResult(split, variance, score, base) =>
+      // Here's the probability rebasing again
+      cumSum = cumSum + score * Math.exp(base - rebase)
+      if (draw < cumSum) {
+        val deltaImpurity = initialVariance - variance
+        return (split, deltaImpurity)
+      }
```
bfolie: Since this logic is repeated in both boltzmann splitters, might want to pull it out as a utility function

src/main/scala/io/citrine/lolo/trees/splits/BoltzmannSplitter.scala
```diff
@@ -0,0 +1,226 @@
...
+      if (draw < cumSum) {
+        return Some(SplitterResult(new RealSplit(index, pivot), variance, totalScore, base))
+      }
+    }
+    // This should never be hit; it would mean there's a bug in the logic above ^^
+    throw new RuntimeException(s"Draw was beyond all the probabilities: ${draw} > $cumSum")
+  }
+
+  /**
+    * Get find the best categorical splitter.
```
bfolie:
```suggestion
    * Find the best categorical splitter.
```

src/test/scala/io/citrine/lolo/AccuracyTest.scala
bfolie: If it's just diagnostic, can it be turned into a test? Is there any expected behavior we can check for?

maxhutch: I'd like to defer maturing it to a follow-on PR.

src/main/scala/io/citrine/lolo/trees/splits/BoltzmannSplitter.scala
maxhutch: Yeah; this language is simply wrong.  I'm revising for correctness and clarity.

src/main/scala/io/citrine/lolo/trees/splits/BoltzmannSplitter.scala
maxhutch: Added

src/main/scala/io/citrine/lolo/trees/splits/BoltzmannSplitter.scala
bfolie: I think that `temperature = 0` (as is allowed here) will cause problems. Because Scala deal with infinities well it won't throw any errors right away, but you'll get `score = 0` for all the splits and `draw = 0`, so you'll end up triggering the error on lines 89/164/237

src/main/scala/io/citrine/lolo/trees/splits/BoltzmannSplitter.scala
```diff
@@ -0,0 +1,239 @@
...
+import scala.collection.mutable
+import scala.util.Random
+
+/**
+  * Find a split for a regression problem
+  *
+  * The splits are picked with a probability that is related to the reduction in variance:
+  * P(split) ~ exp[ - {remaining variance} / ({temperature} * {total variance}) ]
+  * recalling that the "variance" here is weighted by the sample size (so its really the sum of the square difference
+  * from the mean of that side of the split).  This is akin to kinetic monte carlo and simulated annealing techniques.
```
sparadiso: [trivial, nitpick] I might just say "metropolis-hastings" or plain "monte carlo". Kinetic MC highlights an orthogonal consideration (relative rate/event occurrence given fixed amount of elapsed time).

src/main/scala/io/citrine/lolo/trees/splits/BoltzmannSplitter.scala
```diff
@@ -0,0 +1,239 @@
...
+    * @param data         to split
+    * @param numFeatures  to consider, randomly
+    * @param minInstances minimum instances permitted in a post-split partition
+    * @return a split object that optimally divides data
+    */
+  def getBestSplit(data: Seq[(Vector[AnyVal], Double, Double)], numFeatures: Int, minInstances: Int): (Split, Double) = {
+    /* Pre-compute these for the variance calculation */
+    val calculator = VarianceCalculator.build(data.map(_._2), data.map(_._3))
+    val initialVariance = calculator.getImpurity
+    val beta = 1.0 / (temperature * initialVariance)
```
sparadiso: If initialVariance is very small, will other parts of the calculation blow up before (or at the same time) this does?

maxhutch: Good thought.  If its small, its fine.  Zero does cause problems.  I'm adding an early exit and some tests.

PR #96, opened 2017-07-10
Adding 1 ms yield to canStop

maxhutch:
CC: @sparadiso

### DIFF
--- a/src/main/scala/io/citrine/lolo/util/Async.scala
+++ b/src/main/scala/io/citrine/lolo/util/Async.scala
@@ -9,6 +9,7 @@ object Async {
     * Check the thread's interrupted status, because it could stop
     */
   def canStop(): Unit = {
+    Thread.sleep(1) // sleep for 1 ms to yield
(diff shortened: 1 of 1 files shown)


sparadiso:
haha thanks, @maxhutch. I will sleep will tonight!

src/main/scala/io/citrine/lolo/util/Async.scala
```diff
@@ -9,6 +9,7 @@ object Async {
     * Check the thread's interrupted status, because it could stop
     */
   def canStop(): Unit = {
+    Thread.sleep(1) // sleep for 1 ms to yield
```
sparadiso: Why not Thread.yield()?

maxhutch: https://docs.oracle.com/javase/7/docs/api/java/lang/Thread.html#yield()

> "It is rarely appropriate to use this method"

Sleep was suggested as being preferred and far more robust.

PR #99, opened 2017-07-31
Hotfix: duplicate inputs

maxhutch:
The "is this feature value larger than the last one" logic had an absolute, not relative scale, so large feature values might have been mistakenly labeled as distinct.

### DIFF
--- a/src/main/scala/io/citrine/lolo/trees/splits/ClassificationSplitter.scala
+++ b/src/main/scala/io/citrine/lolo/trees/splits/ClassificationSplitter.scala
@@ -55,8 +55,12 @@ object ClassificationSplitter {
         bestSplit = possibleSplit
       }
     }
-    val deltaImpurity = -bestImpurity
-    (bestSplit, deltaImpurity)
+    if (bestImpurity == Double.MaxValue) {
+      (new NoSplit(), 0.0)
+    } else {
+      val deltaImpurity = -bestImpurity
+      (bestSplit, deltaImpurity)
+    }
   }
 
   /**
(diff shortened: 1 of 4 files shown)


src/main/scala/io/citrine/lolo/trees/splits/ClassificationSplitter.scala
```diff
@@ -106,7 +110,8 @@ object ClassificationSplitter {
          1) there is only one branch and
          2) it is usually false
        */
-      if (totalPurity > bestPurity && j + 1 >= minCount && thinData(j + 1)._1 > thinData(j)._1 + 1.0e-9) {
+
+      if (totalPurity > bestPurity && j + 1 >= minCount && Math.abs((thinData(j + 1)._1 - thinData(j)._1)/thinData(j)._1) > 1.0e-9) {
         bestPurity = totalPurity
         /* Try pivots at the midpoints between consecutive member values */
         bestPivot = (thinData(j + 1)._1 + thinData(j)._1) / 2.0 // thinData(j)._1 //
```
sparadiso: Can I make a readability suggestion? Whenever there is a repeated and somewhat dense value lookup or calculation inside of a method, define an inline function to wrap it and give it a meaningful name. 

In this case:

```scala
def featureValueAt(index: Int): Any = thinData(index)._1
```

Or even

```scala
// Look up the feature value at a given index
def x(index: Int): Any = thinData(index)._1
```

sparadiso: (Obviously this would be overridden by runtime considerations in performance critical code, if there would be performance consequences)

PR #107, opened 2017-08-25
Making a small adjustment to uncertainty calculations; adding test

maxhutch:
The change in the bias calculation is minor; it didn't have a large affect but is motivated by the observation that the uncertainty in this PvA calculation is underestimated with the same ratio as the error (due to getting the exact correct answer when the training row is included).

The major addition is a test of uncertainty calibration, which is re-assuring.

### DIFF
--- a/src/test/scala/io/citrine/lolo/bags/BaggerTest.scala
+++ b/src/test/scala/io/citrine/lolo/bags/BaggerTest.scala
@@ -9,6 +9,8 @@ import io.citrine.lolo.trees.regression.RegressionTreeLearner
 import org.junit.Test
 import org.scalatest.Assertions._
 
+import scala.util.Random
+
 /**
   * Created by maxhutch on 11/29/16.
   */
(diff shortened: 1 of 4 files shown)


src/test/scala/io/citrine/lolo/bags/BaggerTest.scala
```diff
@@ -77,6 +77,35 @@ class BaggerTest {
...
+    * by the predicted uncertainty.  On the exterior, these is an additional extrapolative
+    * uncertainty that isn't captured well by this method, so we test the interior and the full
+    * set independently
+    */
+  def testUncertaintyCalibration(): Unit = {
+    val width = 0.1
+    val trainingData = TestUtils.generateTrainingData(128, 6, xscale = width, seed = 6L)
+    val DTLearner = new RegressionTreeLearner(numFeatures = 6)
+    val bias = new RegressionTreeLearner(maxDepth = 4)
+    val baggedLearner = new Bagger(DTLearner, numBags = 16 * trainingData.size, biasLearner = Some(bias))
```
jling11: Why 16?

src/test/scala/io/citrine/lolo/bags/BaggerTest.scala
```diff
@@ -77,6 +77,35 @@ class BaggerTest {
...
+    val RFMeta = baggedLearner.train(trainingData)
+    val RF = RFMeta.getModel()
+
+    val interiorTestSet = TestUtils.generateTrainingData(64, 6, xscale = width/2.0, xoff = width/4.0, seed = 7L)
```
jling11: Why the offset?

src/test/scala/io/citrine/lolo/bags/BaggerTest.scala
```diff
@@ -77,6 +77,35 @@ class BaggerTest {
...
+    val fullTestSet = TestUtils.generateTrainingData(64, 6, xscale = width, seed = 7L)
+
+    val interiorStandardRMSE = BaggerTest.getStandardRMSE(interiorTestSet, RF)
+    val fullStandardRMSE = BaggerTest.getStandardRMSE(fullTestSet, RF)
+    assert(interiorStandardRMSE > 0.85, "Standard RMSE in the interior should be greater than 0.85")
+    assert(interiorStandardRMSE < 1.0, "Standard RMSE in the interior should be less than 1.0")
+
```
jling11: With what probability should these assert statements be true if you don't fix the random seeds?

src/test/scala/io/citrine/lolo/bags/BaggerTest.scala
```diff
@@ -160,6 +189,21 @@ object BaggerTest {
...
+  def getStandardRMSE(testSet: Seq[(Vector[Any], Double)], model: BaggedModel): Double = {
+    val predictions = model.transform(testSet.map(_._1))
+    val pva = testSet.map(_._2).zip(
+      predictions.getExpected().asInstanceOf[Seq[Double]].zip(
+        predictions.getUncertainty().get.asInstanceOf[Seq[Double]]
+      )
+    )
+    val standardError = pva.map{ case (a: Double, (p: Double, u: Double)) =>
+      Math.abs(a - p) / u
```
jling11: Is there a case where u could be equal to zero?  If so, we should set a minimum value for u here.

src/test/scala/io/citrine/lolo/bags/BaggerTest.scala
maxhutch: Large to reduce the influence of the bias correction (which causes over-estimation).  16 speciifcally because it was the smallest value that gave good agreement with 1.0

src/test/scala/io/citrine/lolo/bags/BaggerTest.scala
maxhutch: The training set is on [0, width]^6.  The interior test set is on [width/4, 3*width/4]^6.  The offset pushes it up from [0]

src/test/scala/io/citrine/lolo/bags/BaggerTest.scala
maxhutch: If `u == 0`, its an error.

src/test/scala/io/citrine/lolo/bags/BaggerTest.scala
jling11: I'd pull this magic number out into its own variable, and include a comment to the effect of what you've said above.

PR #108, opened 2017-08-25
Access hypers through a getter, which can pull hypers from internal learners

maxhutch:
This way one were to set a hyperparameter in, say, a leaf learner,
it would be gotten by getHypers called on the parent

### DIFF
--- a/src/main/scala/io/citrine/lolo/bags/Bagger.scala
+++ b/src/main/scala/io/citrine/lolo/bags/Bagger.scala
@@ -25,12 +25,17 @@ class Bagger(
               biasLearner: Option[Learner] = None
             ) extends Learner {
 
+  hypers = Map("useJackknife" -> useJackknife, "numBags" -> numBags)
+
   override def setHypers(moreHypers: Map[String, Any]): Bagger.this.type = {
     method.setHypers(moreHypers)
     super.setHypers(moreHypers)
   }
 
-  override var hypers: Map[String, Any] = Map("useJackknife" -> useJackknife, "numBags" -> numBags)
+  override def getHypers(): Map[String, Any] = {
+    method.getHypers() ++ hypers
+  }
+
 
   private def combineImportance(v1: Option[Vector[Double]], v2: Option[Vector[Double]]): Option[Vector[Double]] = {
     (v1, v2) match {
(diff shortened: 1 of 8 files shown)


src/main/scala/io/citrine/lolo/trees/regression/RegressionTree.scala
```diff
@@ -21,18 +21,21 @@ class RegressionTreeLearner(
                              maxDepth: Int = 30,
                              leafLearner: Option[Learner] = None
                            ) extends Learner {
+  /** Learner to use for training the leaves */
+  val myLeafLearner = leafLearner.getOrElse(new GuessTheMeanLearner())
+
+  /** Hyperparameters */
+  hypers = Map("minLeafInstances" -> 1, "maxDepth" -> maxDepth, "numFeatures" -> numFeatures)
```
sparadiso: Why not use the setter here?

src/main/scala/io/citrine/lolo/bags/Bagger.scala
```diff
@@ -25,12 +25,17 @@ class Bagger(
...
+  hypers = Map("useJackknife" -> useJackknife, "numBags" -> numBags)
+
   override def setHypers(moreHypers: Map[String, Any]): Bagger.this.type = {
     method.setHypers(moreHypers)
     super.setHypers(moreHypers)
   }
 
-  override var hypers: Map[String, Any] = Map("useJackknife" -> useJackknife, "numBags" -> numBags)
+  override def getHypers(): Map[String, Any] = {
+    method.getHypers() ++ hypers
```
sparadiso: noice

src/main/scala/io/citrine/lolo/trees/classification/ClassificationTree.scala
```diff
@@ -19,14 +19,18 @@ class ClassificationTreeLearner(
 
   val myLeafLearner: Learner = leafLearner.getOrElse(new GuessTheMeanLearner)
 
-  override var hypers: Map[String, Any] = Map("maxDepth" -> 30, "minLeafInstances" -> 1, "numFeatures" -> numFeatures)
+  hypers = Map("maxDepth" -> 30, "minLeafInstances" -> 1, "numFeatures" -> numFeatures)
```
sparadiso: setter?

PR #109, opened 2017-08-26
Release: 0.3.0

maxhutch:
Changes:
 - Minor change in the calculation of the bias signal in `Bagger`
 - Better and more consistent handling of hyperparameters, moving them mostly to set and get methods


sparadiso:
☕️

maxhutch:
Its the performance tests.  If the testing nodes are unstable, then it will run the performance tests many times.  I can reduce the maximum number of runs to avoid this.

PR #114, opened 2017-10-04
Feature: MultiTaskTrees

maxhutch:
Adding multitask learners with a first implementation of multitask trees.  A multitask learner emits multiple training results, at which point it follows the same pattern as a single task learner.

### DIFF
--- a/src/main/scala/io/citrine/lolo/trees/multitask/MultiTaskTrainingNode.scala
+++ b/src/main/scala/io/citrine/lolo/trees/multitask/MultiTaskTrainingNode.scala
@@ -1,0 +1,50 @@
+package io.citrine.lolo.trees.multitask
+
+import io.citrine.lolo.PredictionResult
+import io.citrine.lolo.linear.GuessTheMeanLearner
+import io.citrine.lolo.trees.{InternalModelNode, ModelNode, TrainingLeaf}
+import io.citrine.lolo.trees.splits.{MultiTaskSplitter, NoSplit}
+
+class MultiTaskTrainingNode(inputs: Seq[(Vector[AnyVal], Array[AnyVal], Double)]) {
+
+  val (split, _) = MultiTaskSplitter.getBestSplit(inputs, inputs.head._1.size, 1)
+  val (leftChild: Option[MultiTaskTrainingNode], rightChild: Option[MultiTaskTrainingNode]) = split match {
+    case _: NoSplit => (None, None)
+    case _: Any =>
+      val (leftData, rightData) = inputs.partition(row => split.turnLeft(row._1))
+      (Some(new MultiTaskTrainingNode(leftData)), Some(new MultiTaskTrainingNode(rightData)))
+  }
+
+  lazy val leftNodes = leftChild.get.getNodes()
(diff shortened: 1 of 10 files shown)


maxhutch:
@sparadiso @jling11 @eantono, this is ready for a go-over.  I plan to address performance in a follow-up PR.

src/main/scala/io/citrine/lolo/trees/splits/MultiTaskSplitter.scala
```diff
@@ -0,0 +1,166 @@
...
+  /**
+    * Compute the impurity of a set of weighted labels
+    * @param labels is a seq of (Array of multiple labels, single weight)
+    * @return the impurity, which is in [0, number of labels * sum of weights]
+    */
+  def computeImpurity(labels: Seq[(Array[AnyVal], Double)]): Double = {
+    // Return early if there is no impurity
+    if (labels.size == 1 || labels.head._1.isEmpty || labels.map(_._2).sum == 0.0) return 0.0
+
+    // Sum the impurity of each individual label
```
jling11: Is some normalization required if the labels vary over different ranges?

maxhutch: Yes, but I'd like to leave that up to the user.  They can used the standardizer for the labels if they want.  I'll add a multitask standardizer

src/main/scala/io/citrine/lolo/trees/multitask/MultiTaskTrainingNode.scala
```diff
@@ -0,0 +1,66 @@
...
+
+import io.citrine.lolo.PredictionResult
+import io.citrine.lolo.linear.GuessTheMeanLearner
+import io.citrine.lolo.trees.splits.{MultiTaskSplitter, NoSplit}
+import io.citrine.lolo.trees.{InternalModelNode, ModelNode, TrainingLeaf}
+
+/**
+  * Node in a multi-task training tree, which can produce nodes for its model trees
+  *
+  * @param inputs data on which to select spilts and form models
```
sparadiso: splits

src/main/scala/io/citrine/lolo/trees/multitask/MultiTaskTrainingNode.scala
```diff
@@ -0,0 +1,66 @@
...
+  */
+class MultiTaskTrainingNode(inputs: Seq[(Vector[AnyVal], Array[AnyVal], Double)]) {
+
+  // Compute a split
+  val (split, _) = MultiTaskSplitter.getBestSplit(inputs, inputs.head._1.size, 1)
+
+  // Try to construct left and right children
+  val (leftChild: Option[MultiTaskTrainingNode], rightChild: Option[MultiTaskTrainingNode]) = split match {
+    case _: NoSplit => (None, None)
```
sparadiso: `case NoSplit => ` should work here

maxhutch: not a case class :-(

PR #116, opened 2017-10-09
Feature: multitask standardizer

maxhutch:
Just like single task, but with multiple output transformers


maxhutch:
CC: @sparadiso @jling11

PR #117, opened 2017-10-10
Feature: multitask splitter performance

maxhutch:


src/main/scala/io/citrine/lolo/trees/splits/MultiImpurityCalculator.scala
```diff
@@ -0,0 +1,57 @@
+package io.citrine.lolo.trees.splits
+
+class MultiImpurityCalculator(
+                               calculators: Seq[ImpurityCalculator[AnyVal]]
+                             ) extends ImpurityCalculator[Array[AnyVal]] {
+
+  def add(value: Array[AnyVal], weight: Double): Double = {
+    value.zip(calculators).map{case (v, calc) =>
+      if (v.isInstanceOf[Double] && !v.asInstanceOf[Double].isNaN){
```
sparadiso: Can this branch be made earlier (instead of in the inner loop)

maxhutch:
This was reviewed by @sparadiso

PR #130, opened 2018-08-07
Bugfix/nan in get gradient standardizer

astha-chem:
A bugfix to prevent getGradient() from returning 0 x Inf=NaN


src/main/scala/io/citrine/lolo/transformers/Standardizer.scala
```diff
@@ -184,6 +184,8 @@ class StandardizerPrediction[T](baseResult: PredictionResult[T], trans: Seq[Opti
       case None => None
       case Some(x) =>
         Some(x.map(g => g.zip(trans.tail).map {
+          // Return 0.0 if trying to undo inverse of 0.0
+          case (0.0, Some((_, Double.PositiveInfinity))) => 0
```
maxhutch: Having an affine transformation with infinite stretch is probably wrong to begin with.  Could we try to catch it earlier (e.g in the calculation of `trans`)?

astha-chem:
@maxhutch : Added a new test here. I removed the previously proposed bugfix since it does not pass this test. Also feel free to ignore this other PR which is a duplicate (couldn't figure out how to withdraw it): https://github.com/CitrineInformatics/lolo/pull/131

astha-chem:
@maxhutch : I have modified our test testStandardWithConstantFeature such that it now fails without the 0xInf bugfix I had proposed earlier. The fix is currently commented in the code so the test is failing returning "Failed test for gradient. 0.0 and NaN gradients should be the same"

As a side note, I noticed that if I run the test several times, it also fails on 'expected', i.e. expected and standardExpected don't always match. That looks like a separate issue to me, and I have not dug into it. I just moved that assert statement below so that the gradient comparison assert would fail first.

astha-chem:
@maxhutch : I addressed your comments. This is ready.

PR #133, opened 2018-08-20
Bugfix: drop constant features in linear models

maxhutch:
The unregularized linear regression problem is ill posed if there are constant features.  This change skips constant features if there is no regularization (just like we skip NaN and non-real features).


maxhutch:
@astha-chem how does this look?

PR #139, opened 2018-11-26
Add an initial draft of lolopy

WardLT:
Shows training a model and getting uncertainties out.


python/lolopy/learners.py
```diff
@@ -0,0 +1,204 @@
...
+        # Get the model out
+        self.model_ = result.getModel()
+
+        return self
+
+    @abstractmethod
+    def _make_learner(self):
+        """Instantiate the learner used by Lolo to train a model
+
+        Ret"""
```
maxhutch: Is this incomplete?

python/lolopy/learners.py
```diff
@@ -0,0 +1,204 @@
...
+            # Copy the X into an array
+            x_i_java = ListConverter().convert(x_i, self.gateway._gateway_client)
+            x_i_java = self.gateway.jvm.scala.collection.JavaConverters.asScalaBuffer(
+                x_i_java).toVector()
+
+            # If a regression problem, make sure y_i is a float
+            if is_regressor(self):
+                y_i = float(y_i)
+            else:
+                y_i = int(y_i)
```
maxhutch: Does this imply that string labels are not supported?

python/lolopy/learners.py
WardLT: It sure is. Fixing it now

python/lolopy/learners.py
```diff
@@ -0,0 +1,204 @@
...
+
+            # If weights are None, RF expects a 3ple
+            if weights is None:
+                pair = self.gateway.jvm.scala.Tuple3(x_i_java, y_i, 1.0)
```
maxhutch: Why not incorporate the weights here?

python/lolopy/learners.py
```diff
@@ -0,0 +1,204 @@
...
+    def fit(self, X, y, weights=None):
+        # Get the number of classes
+        self.n_classes_ = len(set(y))
+
+        return super(RandomForestClassifier, self).fit(X, y, weights)
+
+    def predict(self, X):
+        return super(RandomForestClassifier, self).predict(X, False)
+
+    def predict_proba(self, X):
```
maxhutch: `predict_proba`?  Did you mean `predict_probability`?

python/setup.py
```diff
@@ -0,0 +1,8 @@
+from setuptools import setup
+
+setup(
+    name='lolopy',
+    version='0.0.1',
+    packages=['lolopy'],
+    description='Python wrapper for the Lolo machine learning library',
```
maxhutch: We should add sklearn and py4j to install_requirements.

python/lolopy/learners.py
WardLT: No, `predict_proba` is the proper function name for scikit-learn classifiers.

python/lolopy/learners.py
WardLT: That is correct. I need the class labels to have a defined ordering, and the easiest route to that is to use integers and employ the [LabelEncoder](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.LabelEncoder.html) built in to ScikitLearn for strings, etc.

python/setup.py
WardLT: Done

PR #141, opened 2018-12-27
Enable lolopy profile on deploy

maxhutch:
This will deploy the jar-with-dependencies next to the normal jar.
It also cleans up the deploy process to use some newer travis
config options.


maxhutch:
We won't really be able to test this til we get back to the master branch.

PR #142, opened 2018-12-28
Working Draft of Lolopy

WardLT:
Significantly revised lolopy:

- Refactored the base class to make adding new learners easier
- Improved performance of data transfer between Python and JVM
- Added examples and scaling tests
- Added installation instructions
- Modified Travis CI build script

The major to-do left on this implementation is a method for downloading the lolo jar file, if needed. Currently, we hard-code the path to find the appropriate jar in the `./target/` directory.


.travis.yml
```diff
@@ -1,17 +1,30 @@
...
+        - "~/.m2/repository"
+      after_success:
+      - "./cd/before-deploy.sh"
+      - "./cd/deploy.sh"
+    - language: "python"
+      python:
+        - 3.6
+      before_script: cd python
+      install:
+        - make
```
maxhutch: this is being invoked from the root dir but the makefile is in `python/`.  You could do `make python/Makefile`

python/README.md
```diff
@@ -1,3 +1,58 @@
...
+```
+
+The results of this code is to produce the predicted values (`y_pred`) and their uncertainties (`y_std`).
+
+See the [`examples`](./examples) folder for more examples and details.
+
+## Implementation and Performance
+
+`lolopy` is built using the [Py4J](https://www.py4j.org/) library to interface with the Lolo scala library.
+Py4J provides the ability to easily managing a JVM server, create Java objects in that JVM, and calling Java methods from Python.
```
maxhutch: Grammar: `ability to easily manage...and call Java methods from Python.`

python/README.md
```diff
@@ -1,3 +1,58 @@
...
+## Implementation and Performance
+
+`lolopy` is built using the [Py4J](https://www.py4j.org/) library to interface with the Lolo scala library.
+Py4J provides the ability to easily managing a JVM server, create Java objects in that JVM, and calling Java methods from Python. 
+However, Py4J  [has slow performance in transfering large arrays](https://github.com/bartdag/py4j/issues/159).
+To transfer arrays of features (e.g., training data) to the JVM before model training or evaluation, we transform the data to/from Byte arrays on the Java and Python sides. 
+Transfering data as byte arrays does allow for quickly moving data between the JVM and Python but requires holding 3 copies of the data in memory at once (Python, Java Byte array, and Java numerical array).
+We could reduce memory usage by passing the byte array in chunks, but this is currently not implemented.
+
+Our performance for modeling training is comparable to scikit-learn, as shown in the figure below.
```
maxhutch: Grammar: `model training`

python/lolopy/learners.py
```diff
@@ -137,39 +125,52 @@ def predict(self, X, return_std=False):
         # Get the PredictionResult
         pred_result = self.model_.transform(X_java)
 
+        # Unlink the run data, which is no longer needed (to save memory)
+        self.gateway.detach(X_java)
+
         # Pull out the expected values
-        exp_values = pred_result.getExpected()
-        y_pred = [exp_values.apply(i) for i in range(len(X))]
+        y_pred = self.gateway.jvm.io.citrine.lolo.util.LoloPyDataLoader.getRegressionExpected(pred_result)
```
maxhutch: Could you pick a different name to avoid `x = f(x)` type patterns.  Maybe `y_pred_byte`?

python/lolopy/learners.py
```diff
@@ -137,39 +125,52 @@ def predict(self, X, return_std=False):
...
 
         # If desired, return the uncertainty too
         if return_std:
             # TODO: This part fails on Windows because the NativeSystemBLAS is not found. Fix that
             # TODO: This is only valid for regression models. Perhaps make a "LoloRegressor" class
-            uncertain = pred_result.getUncertainty().get()
-            y_std = np.zeros_like(y_pred)
-            for i in range(len(X)):
-                y_std[i] = uncertain.apply(i)
+            y_std = self.gateway.jvm.io.citrine.lolo.util.LoloPyDataLoader.getRegressionUncertainty(pred_result)
```
maxhutch: Likewise here

src/main/scala/io/citrine/lolo/util/LoloPyDataLoader.scala
```diff
@@ -0,0 +1,126 @@
...
+    val buffer = ByteBuffer.wrap(input).order(ordering)
+
+    // Generate digits
+    val nDigits = input.length / 8
+    (0 until nDigits).map(x => buffer.getDouble(x * 8)).grouped(numAttributes).map(_.toVector).toVector
+  }
+
+
+  /**
+    * Receive a class or weight array
```
maxhutch: I don't follow what a `class array` is here.

src/main/scala/io/citrine/lolo/util/LoloPyDataLoader.scala
```diff
@@ -0,0 +1,126 @@
...
+    * @param input Input byte array
+    * @param getFloat Whether to read doubles (or integers)
```
maxhutch: how about `getDouble` since they are doubles?

src/main/scala/io/citrine/lolo/util/LoloPyDataLoader.scala
```diff
@@ -0,0 +1,126 @@
...
+    }
+  }
+
+  /**
+    * Zips the features and labels together
+    * @param X Feature array
+    * @param y Label array
+    * @return Zipped arrays
+    */
+  def zipTrainingData(X: Seq[Vector[Double]], y: Seq[Any]) : Seq[(Vector[Double], Any)] = {
```
maxhutch: Does this need to be stand-alone?

src/test/scala/io/citrine/lolo/util/LoloPyDataLoaderTest.scala
```diff
@@ -0,0 +1,111 @@
...
+    stream.read(byteArray)
+    byteArray
+  }
+
+  @Test
+  def testFeatureArray(): Unit = {
+    val data = LoloPyDataLoader.getFeatureArray(getData, 4, bigEndian = false)
+    assert(data.length == 4)
+    assert(data.head.length == 4)
+    assert(Math.abs(data.head(1) - 1) < 1e-6, "Array read in wrong order")
```
maxhutch: Could you test the other values?  This test is insensitive to transposes.

src/test/scala/io/citrine/lolo/util/LoloPyDataLoaderTest.scala
```diff
@@ -0,0 +1,111 @@
...
+  @Test
+  def test1DArrayInteger(): Unit = {
+    val data = LoloPyDataLoader.get1DArray(getData, getFloat = false, bigEndian = false).asInstanceOf[Seq[Int]]
+    assert(data.length == 32)
+    assert(data.take(3).sum == 0) // First 3 values are all 0
+    assert(data(3) == 1072693248) // Fourth is a big one
+  }
+
+  @Test
+  def testZipping(): Unit = {
```
maxhutch: Does this need to be tested?

src/test/scala/io/citrine/lolo/util/LoloPyDataLoaderTest.scala
```diff
@@ -0,0 +1,111 @@
...
+      TestUtils.generateTrainingData(32, 12, noise = 0.1, function = Friedman.friedmanSilverman),
+      inputBins = Seq((0, 8))
+    )
+
+    val RFMeta = RandomForest.apply(numTrees = trainingData.length * 2)
+      .train(trainingData)
+    val RF = RFMeta.getModel()
+
+    // Generate the prediction results object
+    val results = RF.transform(trainingData.map(_._1))
```
maxhutch: Why do the results need to come from a model?  Couldn't you just generate some random means and sigmas and then round-robin them through?

src/test/scala/io/citrine/lolo/util/LoloPyDataLoaderTest.scala
```diff
@@ -0,0 +1,111 @@
...
+      TestUtils.generateTrainingData(32, 12, noise = 0.1, function = Friedman.friedmanSilverman),
+      inputBins = Seq((0, 8)), responseBins = Some(8)
+    )
+    val RFMeta = new RandomForest(numTrees = trainingData.length * 2)
+      .train(trainingData.map(x => (x._1, Integer.parseInt(x._2.toString))))
+    val RF = RFMeta.getModel()
+    val nClasses : Int = trainingData.map(x => Integer.parseInt(x._2.toString)).max + 1
+
+    // Get the base results
+    val results = RF.transform(trainingData.map(_._1))
```
maxhutch: Why do the results need to come from a model?  Couldn't you just generate some random means and probs and then round-robin them through?

src/test/scala/io/citrine/lolo/util/LoloPyDataLoaderTest.scala
WardLT: This test should be sensitives to transposes because the array is [[0, 1, 2, 3], [4, ...]]. But, for my own paranoia's sake, I'll add in another check

src/main/scala/io/citrine/lolo/util/LoloPyDataLoader.scala
WardLT: Yep, that makes sense. Fix on the way

src/main/scala/io/citrine/lolo/util/LoloPyDataLoader.scala
WardLT: I'm changing it to 1D array to be more generic. What I meant by "class array" was "array holding the values of the class variable (i.e., the one to be predicted by the ML model)"

WardLT:
This PR is ready for your review again. I've made all of your requested changes, with the exception of the "Why do the results need to come from a model?" where I'm not sure what to do.

python/lolopy/learners.py
WardLT: Just realized I missed this one

src/test/scala/io/citrine/lolo/util/LoloPyDataLoaderTest.scala
maxhutch: Yes.  You can create an anonymous class that extends `PredictionResult` (https://stackoverflow.com/questions/24657450/scala-how-to-use-extends-with-an-anonymous-class)

src/test/scala/io/citrine/lolo/util/LoloPyDataLoaderTest.scala
maxhutch: My suggestion would be to remove it, but I'd ask that you at least comment that it is just testing zip and not something more subtle.  If I were reading this code, I'd probably spend a while trying to figure out what it was testing since I'd assume it wasn't just testing zip.

src/main/scala/io/citrine/lolo/util/LoloPyDataLoader.scala
maxhutch: Could you mention that's why you made it in the comment?

src/test/scala/io/citrine/lolo/util/LoloPyDataLoaderTest.scala
WardLT: Makes sense. I'll remove it.

src/test/scala/io/citrine/lolo/util/LoloPyDataLoaderTest.scala
WardLT: Cool. I've done that for both of the test where I trained a model

src/main/scala/io/citrine/lolo/util/LoloPyDataLoader.scala
WardLT: Yep, that's a good idea. I've added an explanation

maxhutch:
LGTM

PR #147, opened 2019-01-08
Prepare Lolo for Publication to PyPi

WardLT:
This PR adds several features needed before publication to PyPi:

- Copies the jar file from lolo to a directory inside of the package when building for PyPi
- Use this version of the lolo if lolo is not installed in a development environment
- Added tests for using either the package or development version of the lolo jar file
- Updated instructions for installation
- Added a better way of controlling the version number


WardLT:
Whoops, added the commits twice due to some bad merging. 

Going to fix that.

python/setup.py
```diff
@@ -1,9 +1,56 @@
 from setuptools import setup
+from glob import glob
+import shutil
+import sys
+import os
 
+# single source of truth for package version
+version_ns = {}
+with open(os.path.join("lolopy", "version.py")) as f:
+    exec(f.read(), version_ns)
```
maxhutch: I'm not familiar with this pattern; why are you using `exec`?

WardLT: I was following one of these [patterns for accessing the library version from `setup.py`](https://packaging.python.org/guides/single-sourcing-package-version/).

If `exec` makes you nervous, there are other routes (e.g., extracting the version number with a regex) .

WardLT:
On further thought, we might not be able to use the backed-in pypi deployment script to travis because the setup.py is not in the root directory. So, looks like we'd need our own script

python/setup.py
maxhutch: No, the pattern is fine.  Thanks for the link.

maxhutch:
Writing our own script is fine.  I would like to have CD; I don't consider forcing a lolo version update for a lolopy update to be a show-stopper.  That's what patch versions are for :-D

Do you want to take a crack at that script in this PR, or push it to another (and potentially to me)?

WardLT:
Sure, I can take a crack at the deployment script.

.travis.yml
```diff
@@ -20,11 +25,15 @@ matrix:
         - pip install --only-binary=numpy,scipy -r requirements.txt
       script:
         - nosetests
-deploy:
-  provider: script
-  script: bash cd/deploy.sh
-  on:
-    branch: master
+      deploy:
```
maxhutch: How does this jive with "we might not be able to use the backed-in pypi deployment script to travis because the setup.py is not in the root directory."

WardLT: It doesn't. Yesterday, I didn't consider having two different deploy statements: one for the scala code and one for python. That lets me execute the deployment from the two different directories.

WardLT: We'll see if I need to add a:

```
before_deploy:
  - cd python
```

to the python build route. I don't think we do, but the deployment not finding setup.py is the error that would not surprise me.

WardLT:
Fingers crossed :)

PR #149, opened 2019-01-17
Statistical validation methods, e.g. cross-validation

maxhutch:
There are chunks of changes here:
 - `Merit`: Figures of merit for model quality.  Given multiple predicted-vs-actual batches, the the error in their estimates is computed as well.  There is also a driver to sweep them across a parameter, e.g. number of trees.
 - `Visualization`: like merits, they take predicted-vs-actual data and assess the quality of a model, but do so via a visual output rather than a scalar.
 - `CrossValidation`: drivers to produce predicted-vs-actual data via cross-validation, specifically k-fold cross-validation with multiple trials.
 - `StatisticalValidation`: drivers to produce predicted-vs-actual data by drawing training and test sets from known distributions.  These are more helpful when studying asymptotic properties of learners.


src/main/scala/io/citrine/lolo/validation/Metric.scala
```diff
@@ -0,0 +1,212 @@
...
+  /**
+    * Compute metrics as a function of a parameter, given a builder that takes the parameter to predicted-vs-actual data
+    * @param parameterName name of the parameter that's being scanned over
+    * @param parameterValues values of the parameter to try
+    * @param metrics to apply at each parameter value
+    * @param logScale whether the parameters should be plotted on a log scale
+    * @param pvaBuilder function that takes the parameter to predicted-vs-actual data
+    * @return an [[XYChart]] that plots the metrics vs the parameter value
+    */
+  def scanMetrics[T](
```
latture: It was not intuitive that this would produce a chart. I would've expected `scanMetrics` to generate results (of some kind). Then I would expect to use `visualize` to generate the plot. I'm not sure it would be better with the proposed way (since you ultimately want the plot). Just flagging it as something that seemed odd.

src/main/scala/io/citrine/lolo/validation/Visualization.scala
```diff
@@ -0,0 +1,133 @@
...
+    */
+  def visualize(data: Iterable[(PredictionResult[T], Seq[T])]): Chart[_, _]
+}
+
+/**
+  * Histogram of the error divided by the predicted uncertainty
+  * @param nBins number of bins in the histogram
+  * @param range of the horizontal axis, e.g. x \in [-range/2, range/2]
+  */
+case class StandardResidualHistogram(nBins: Int = 128, range: Double = 8.0) extends Visualization[Double] {
```
latture: I can imagine the number of these visualization classes could get quite large. 

At this point I would argue what I'm about to propose isn't worth it, but I wonder if we could create a more compassable API for generating plots that wouldn't require someone to define a `Visualization` class from scratch every time they wanted to make a small tweak.

maxhutch: I think "tweak"-like changes should be codified in the state of the `Visualization`, e.g. `nBins`.  The charts that are emitted are also editable, so a user could "compose" a new one by building another chart inside of it and then tweaking it before returning it.

src/main/scala/io/citrine/lolo/validation/Metric.scala
maxhutch: Good point; renamed.

src/main/scala/io/citrine/lolo/validation/Visualization.scala
latture: True. Tweaking the plot that's generated could cover most use cases. This solution is fine with me.

latture:
Skipping tests for plots seems reasonable to me.

WardLT:
I agree, skipping those tests seems fine.

maxhutch:
@latture @WardLT this is ready for a final review.  One wide-reaching change since you last reviewed is renaming `Metric` to `Merit` to avoid confusion with https://en.wikipedia.org/wiki/Metric_(mathematics) (which we will probably add if/when we add kernel methods).  h/t to @brandonwillard for pointing out the collision.

src/main/scala/io/citrine/lolo/validation/Merit.scala
```diff
@@ -0,0 +1,215 @@
...
+import java.util
+
+import io.citrine.lolo.PredictionResult
+import org.knowm.xchart.XYChart
+
+import scala.collection.JavaConverters._
+import scala.util.Random
+
+/**
+  * Real-valued metric on predictions of type T
```
latture: It may be worth changing `metric` to figure of merit in the scaladoc as well.

src/main/scala/io/citrine/lolo/validation/Merit.scala
```diff
@@ -0,0 +1,215 @@
...
+      (actual + error, uncertainty, actual)
+    }
+
+    /*
+    val sigmaMean = pua.map(_._2).sum / pua.size
+    val sigmaVar = pua.map(x => Math.pow(x._2 - sigmaMean, 2)).sum / pua.size
+    val model = Math.sqrt(2 / (Math.PI - 2)) * Math.sqrt(sigmaVar) / sigmaMean
+    val estimate = computeFromPredictedUncertaintyActual(baseline)
+    // println(model, estimate)
+    */
```
latture: Did you intend to leave this commented block of code?

src/main/scala/io/citrine/lolo/validation/Merit.scala
```diff
@@ -0,0 +1,215 @@
...
+
+    computeFromPredictedUncertaintyActual(predictedUncertaintyActual) / computeFromPredictedUncertaintyActual(ideal)
+  }
+
+  /**
+    * Covariance(X, Y) / Sqrt(Var(X) * Var(Y)), where X is predicted uncertainty and Y is magnitude of error
+    */
+  def computeFromPredictedUncertaintyActual(
+                                             pua: Seq[(Double, Double, Double)]
```
latture: Could we document `@param pua` in the scaladoc?  Maybe something like:
```
@param pua A sequence of tuples. Each tuple contains predicted value, uncertainty and actual value, respectively.
```

src/test/scala/io/citrine/lolo/stats/functions/LinearTest.scala
```diff
@@ -0,0 +1,13 @@
+package io.citrine.lolo.stats.functions
+
+import org.junit.Test
+
+class LinearTest {
+
+  @Test
+  def testSimpleLinear(): Unit = {
+    val func: Linear = Linear(Seq(1.0))
+    println(func(Seq(1.0, 2.0, 3.0)))
```
latture: Did you mean to replace the `println` with an `assert`?

src/test/scala/io/citrine/lolo/validation/MeritTest.scala
```diff
@@ -0,0 +1,158 @@
...
+  def testZeroUncertaintyCorrelation(): Unit = {
+    val pva = getNormalPVA(noiseScale = 0.01, uncertaintyCorrelation = 0.0, batchSize = 256, numBatch = 32)
+    val expected = 0.0
+    val (corr, uncertainty) = UncertaintyCorrelation.estimate(pva)
+    assert(Math.abs(corr - expected) < 3 * uncertainty, "Uncertainty correlation estimate was not accurate enough")
+    assert(uncertainty < 0.05, s"Uncertainty correlation estimate was not precise enough")
+  }
+}
+
+object MeritTest {
```
latture: More of a general question: should objects like this stay with the unit tests? (And, are they useful after initial prototyping?)

src/main/scala/io/citrine/lolo/validation/Merit.scala
```diff
@@ -0,0 +1,215 @@
...
+  def estimate(pva: Iterable[(PredictionResult[T], Seq[T])]): (Double, Double) = {
+    val samples = pva.map { case (prediction, actual) => evaluate(prediction, actual) }
+    val mean: Double = samples.sum / samples.size
+    val variance: Double = (samples.size / (samples.size - 1)) * samples.map(x => Math.pow(x - mean, 2)).sum / samples.size
+    (mean, Math.sqrt(variance / samples.size))
+  }
+}
+
+/**
+  * SSIA
```
latture: I think even a simple description would be helpful here. 

```suggestion
  * Square root of the mean square error. For an unbiased estimator, this is equal to the standard deviation of the difference between predicted and actual values.
```

(If I understand this figure of merit correctly, but you probably have a better explanation.)

src/test/scala/io/citrine/lolo/validation/CrossValidationTest.scala
```diff
@@ -0,0 +1,34 @@
...
+
+    val trainingResult = learner.train(data)
+    val rmseFromPVA = Math.sqrt(
+      trainingResult.getPredictedVsActual().get.map {
+        case (_, p: Double, a: Double) => Math.pow(p - a, 2.0)
+      }.sum / trainingResult.getPredictedVsActual().get.size
+    )
+
+    // These have a false negative rate less than 1/100 at the time of original authorship
+    assert(rmseFromPVA < rmseFromCV + uncertainty)
```
WardLT: Are there ways of setting the random seed for this test to avoid it randomly failing? [I'm assuming that the chances are low, so this isn't a comment to hold up the PR on.

WardLT:
Looks good to me too.

src/test/scala/io/citrine/lolo/validation/MeritTest.scala
maxhutch: I think this will be useful whenever someone adds a unit test, so I'm inclined to keep it.

src/test/scala/io/citrine/lolo/validation/CrossValidationTest.scala
maxhutch: Not currently.  Its complicated by the use of scala's parallel collections, which will access random state in non-deterministic orders.

PR #151, opened 2019-01-24
Added merits/metrics and more learners to lolopy

WardLT:
Adds the ability to use Lolo merits from lolopy

I had to add a new utility interface to lolo to allow for making a PredictionResult from Python data. So, we will need to up the version of the lolo package when we eventually sync develop with master.

I am sending this PR now so that I can use Travis.

### DIFF
--- a/python/lolopy/learners.py
+++ b/python/lolopy/learners.py
@@ -4,6 +4,7 @@ import numpy as np
 from lolopy.loloserver import get_java_gateway
 from lolopy.utils import send_feature_array, send_1D_array
 from sklearn.base import BaseEstimator, RegressorMixin, ClassifierMixin, is_regressor
+from sklearn.exceptions import NotFittedError
 
 __all__ = ['RandomForestRegressor', 'RandomForestClassifier']
 
@@ -147,13 +148,34 @@ class BaseLoloLearner(BaseEstimator, metaclass=ABCMeta):
 
         return self.gateway.jvm.io.citrine.lolo.util.LoloPyDataLoader.getFeatureArray(X.tobytes(), X.shape[1], False)
 
+    def get_importance_scores(self, X):
+        """Get the importance scores for each entry in the training set for each prediction
+        
+        Args:
+            X (ndarray): Inputs for each entry to be assessed
+        """
 
-class BaseLoloRegressor(BaseLoloLearner, RegressorMixin):
-    """Abstract class for models that produce regression models.
+        pred_result = self._get_prediction_result(X)
 
-    Implements the predict operation"""
+        y_import_bytes = self.gateway.jvm.io.citrine.lolo.util.LoloPyDataLoader.getImportanceScores(pred_result)
+        y_import = np.frombuffer(y_import_bytes, 'float').reshape(len(X), -1)
+        return y_import
 
-    def predict(self, X, return_std=False):
+    def _get_prediction_result(self, X):
+        """Get the PredictionResult from the lolo JVM
+        
+        The PredictionResult class holds methods that will generate the expected predictions, uncertainty intervals, etc
+        
+        Args:
+            X (ndarray): Input features for each entry
+        Returns:
+            (JavaObject): Prediction result produced by evaluating the model
+        """
+
+        # Check that the model is fitted
+        if self.model_ is None:
+            raise NotFittedError()
+
         # Convert the data to Java
         X_java = self._convert_run_data(X)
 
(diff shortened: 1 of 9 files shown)


maxhutch:
Is there a metrics interface in sklearn, or is it just a pattern?

maxhutch:
> I am sending this PR now so that I can use Travis.

If that implies that its not ready for a review, can you add the "not yet" label?

WardLT:
Being functions, they don't have an abstract class or anything.

WardLT:
The metrics are done, and I've started in to making other features.

So, this is ready for review to sync the develop branch with my work.

WardLT:
Well, I'll fix the tests first

maxhutch:
@WardLT can you update to the latest `develop`?

WardLT:
Done. We'll see if it passes tests :)

python/lolopy/learners.py
```diff
@@ -83,6 +85,11 @@ def fit(self, X, y, weights=None):
         # Get the model out
         self.model_ = result.getModel()
 
+        # Store the feature importances
+        feature_importances_java = result.getFeatureImportance().get()
+        feature_importances_bytes = self.gateway.jvm.io.citrine.lolo.util.LoloPyDataLoader.send1DArray(feature_importances_java)
+        self.feature_importances_ = np.frombuffer(feature_importances_bytes, 'float')
```
maxhutch: Does sklearn define a getter for the feature importances? If so, could you add a test that accesses it?

python/lolopy/tests/test_learners.py
```diff
@@ -27,6 +53,9 @@ def test_rf_regressor(self):
         # Test with weights (make sure it doesn't crash)
         rf.fit(X, y, [1.0]*len(y))
 
+        # Make sure feature importances are stored
+        self.assertEqual(np.shape(rf.feature_importances_), (X.shape[1],))
```
maxhutch: Could you also test that they sum to 1?

src/main/scala/io/citrine/lolo/util/LoloPyDataLoader.scala
```diff
@@ -163,4 +179,17 @@ object LoloPyDataLoader {
...
+  /**
+    * Create a PredictionResult object from the mean and uncertainty
+    * @param expected Mean of the predictions of a model
+    * @param uncertainty Uncertainty of the predictions
+    * @return Prediction result object
+    */
+  def makeRegressionPredictionResult(expected: Seq[Double], uncertainty: Seq[Double]) : PredictionResult[Double] = {
+    new PredictionResult[Double] {
+      override def getExpected(): Seq[Double] = expected
+      override def getUncertainty(): Option[Seq[Any]] = {new Some(uncertainty)}
```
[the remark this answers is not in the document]
WardLT: Yes, it can. I'll make the change

python/lolopy/learners.py
WardLT: [sklearn does define a getter](https://github.com/scikit-learn/scikit-learn/blob/7389dbac82d362f296dc2746f10e43ffa1615660/sklearn/ensemble/forest.py#L365) because they only compute importances on-demand. I always store them (to avoid having to store the entire training result class), so don't use a getter pattern. That said, I could implement one so we get "model not fitted" errors just like sklearn.

python/lolopy/tests/test_learners.py
WardLT: Yep, will be in the next commit

python/lolopy/tests/test_learners.py
```diff
@@ -1,24 +1,50 @@
...
         y_pred = rf.predict(X)
         self.assertEqual(len(y_pred), len(y))
 
+        # Test the ability to get importane scores
+        y_import = rf.get_importance_scores(X[:100, :])
+        self.assertEqual((100, len(X)), y_import.shape)
+
         # Basic test for functionality. R^2 above 0.98 was measured on 27Dec18
         score = r2_score(y_pred, y)
         print('R^2:', score)
```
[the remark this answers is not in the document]
WardLT: Done. Caught another misspelling too in another method

python/lolopy/tests/test_metrics.py
```diff
@@ -0,0 +1,38 @@
...
+
+            # Make the errors and uncertainties
+            draw = multivariate_normal([0, 0], [[1, expected], [expected, 1]], sample_size)
+
+            # Add the errors, and separate out the standard deviations
+            y_pred = y_true + [d[0] * normal(0, 1) for d in draw]
+            y_std = [abs(d[1]) for d in draw]
+
+            # Test with a very large tolerance for now
+            self.assertTrue(abs(uncertainty_correlation(y_true, y_pred, y_std) - expected) < 0.5,
```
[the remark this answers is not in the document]
WardLT: I bumped it up to 32k, which makes the error something like ~0.23 for the expected at 0.75. I don't go much lower even if I increase the number of samples.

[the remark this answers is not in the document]
WardLT: Done, much prettier now :D

WardLT:
I've made the requested changes. Thanks!

python/lolopy/learners.py
maxhutch: With the code as it is, how is a user supposed to access the feature importances?

WardLT: They are available as the `.feature_importances_` attribute, which mirrors sklearn's RandomForest and is something I have failed to document

maxhutch: Weird. (https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.RandomForestRegressor.html#sklearn.ensemble.RandomForestRegressor.feature_importances_)  :man_shrugging:

PR #152, opened 2019-02-07
Re-calibration of bagger uncertainty estimates

maxhutch:
When the size of the ensemble is small, the predicted uncertainty is far too large, causing the overwhelming majority of the predictions to fall within the predicted uncertainty.  Fortunately (and possibly surprisingly), the shape of the (error / uncertainty) distribution is still normal-ish, so calibrated uncertainty estimates can be recovered by simple rescaling.  This is performed on the out-of-bag (error / uncertainty).

The out-of-bag recalibration takes ~3x longer than normal training, but allows the ensemble to produce reasonable uncertainty estimates with only ~64 learners (on a specific test problem).  This reduces the complexity of both train and predict by a linear order in the number of training points, making the majority of modeling tasks much more efficient.

## Mechanical changes
 * An `uncertaintyCalibration` option has been added to `Bagger` with default `false`.  When true, the out-of-bag re-calibration described above is performed
 * The `BaggedResult` has been converted to a `trait` with two implementations: `BaggedSingleResult` and `BaggedMultiResult`.  The `BaggedSingleResult` is used for the out-of-bag re-calibration, and avoids setting up the large matrices used in `BaggedMultiResult`.  It is also a little easier to read.  There is an explicit test for consistency between the two instances of `BaggedResult`.  That revealed a relatively minor bug, which has been corrected.
 * Supporting materials for the calibration study are produced by `CalibrationStudy`.  It is in the test sources to avoid fattening the distribution.
 * Minor changes were made to the validation routines to allow more data to be garbage collected.


maxhutch:
cc: @WardLT @bfolie @eddotman @eantono

PR #153, opened 2019-02-07
Adding strict test that uncertainty should be positive

maxhutch:
Zero predicted uncertainty is an error state; the uncertainty
might be small but it should always be finite.

### DIFF
--- a/src/main/scala/io/citrine/lolo/bags/Bagger.scala
+++ b/src/main/scala/io/citrine/lolo/bags/Bagger.scala
@@ -347,7 +347,7 @@ class BaggedResult(
       val variancePerRow: DenseVector[Double] = 0.5 * (arg(::, i) - Math.E * correction)
 
       /* Impose a floor in case any of the variances are negative (hacked to work in breeze) */
-      val floor: Double = Math.min(0, -min(variancePerRow))
+      val floor: Double = Math.max(0, -min(variancePerRow))
       val rezero: DenseVector[Double] = variancePerRow - floor
       0.5 * (rezero + abs(rezero)) + floor
     }.map(_.toScalaVector())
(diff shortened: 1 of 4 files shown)


src/main/scala/io/citrine/lolo/bags/Bagger.scala
```diff
@@ -347,7 +347,7 @@ class BaggedResult(
       val variancePerRow: DenseVector[Double] = 0.5 * (arg(::, i) - Math.E * correction)
 
       /* Impose a floor in case any of the variances are negative (hacked to work in breeze) */
-      val floor: Double = Math.min(0, -min(variancePerRow))
+      val floor: Double = Math.max(0, -min(variancePerRow))
       val rezero: DenseVector[Double] = variancePerRow - floor
       0.5 * (rezero + abs(rezero)) + floor
```
eantono: I don't really understand this code, but `min` is almost certainly wrong

why `abs(rezero)`? don't likes 350 and 351 guarantee that all elements of rezero are >= 0?

in the case of having a negative variance, the minimum value for the variance will be the absolute value of the largest magnitude negative variance. why is that a reasonable minimum value?

if these things are not important, feel free to ignore

maxhutch: This code is complicated because we're trying to shove everything into vector form for efficiency.  The clearest path to checking this is by looking at the non-vector version [here](https://github.com/CitrineInformatics/lolo/pull/152/files#diff-7c5cef0ff9e4dba1146bc4ff6c4ea36dR123), which is tested against the vector version explicitly [here](https://github.com/CitrineInformatics/lolo/pull/152/files#diff-f15979d127059e3246bfc35f1a3794adR38).  I just didn't want to block the bugfix on the [recalibration PR](https://github.com/CitrineInformatics/lolo/pull/152), since that will take longer to review and will cause other changes.

sparadiso: Ah, that helps @maxhutch. I was also struggling to follow.

PR #157, opened 2019-02-19
Adding a test that fails for small forests with a bias learner

maxhutch:
When there are no bags in the ensemble that have zero instances
of any particularly training row, the jackknife-after-bootstrap
breaks for lack of an out-of-bag estimate.  We protect against this
in two ways:
 * In the Bagger, re-drawing the bootstrap until each training
   instance is missing from at least one bag
 * In the BaggedSingleResult, ignoring jackknife-after-bootstrap
   and only using infinitesimal jackknife if there are no
   out-of-bag predictions

### DIFF
--- a/src/main/scala/io/citrine/lolo/bags/Bagger.scala
+++ b/src/main/scala/io/citrine/lolo/bags/Bagger.scala
@@ -45,6 +45,11 @@ case class Bagger(
     } else {
       trainingData.size
     }
+
+    // We need enough bags such that the probability that the poisson draw is "valid" is at least 50%
+    // Valid here means that for each training point, there is at least one tree that doesn't include it
+    // The probability that the weights are valid is:
+    // (1 - [(1 - 1/e)^{number of trees}])^{number of training points}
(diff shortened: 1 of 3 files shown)


src/main/scala/io/citrine/lolo/bags/Bagger.scala
```diff
@@ -45,12 +45,23 @@ case class Bagger(
     } else {
       trainingData.size
     }
+    val minBags = Math.log(1 - Math.pow(2, -1.0 / trainingData.size)) / Math.log((Math.E - 1) / Math.E)
```
sparadiso: Can you add an inline explanation for this?

maxhutch: Added

maxhutch:
I ran into this when running some downstream tests.  There was a unit test gap here where the small-ensemble tests didn't use the out-of-bag estimates in `Bagger` _and_ there was a bug in the `BaggedMultiResult` that avoided the `NaN`.

I didn't mean to remove the `0 until 16384`; I pattern matched it as calibration, but its actually part of the test.  Re-added.

PR #161, opened 2019-02-27
Adding option to randomize pivot location in decision trees

maxhutch:
The pivot is generally chosen to be the midpoint between the two nearest data points
in the split (i.e. the rightmost point in the left partition and the leftmost point in
the right partition).  If randomizePivotLocation is set, then the pivot location is
drawn uniformaly from the line between those two points.  This can improve
modeling performance and uncertainty quantification in ensemble models, e.g.
random forests.


maxhutch:
cc: @WardLT @mrupp-citrine

maxhutch:
Note: the defaults here are all explicit and `false` for backward compatibility.

PR #162, opened 2019-02-28
Fixed test bug, and polished RF docs

WardLT:
This fixes the bug observed when testing the bias learner. I made the random forest in that test fit to the data less strongly, which should make the bias learner more effective. 

I also made some minor changes to the RF docs

### DIFF
--- a/python/lolopy/learners.py
+++ b/python/lolopy/learners.py
@@ -262,7 +262,7 @@ class RandomForestMixin(BaseLoloLearner):
             leaf_learner (BaseLoloLearner): Learner used at each leaf of the random forest (default: GuessTheMean)
             subset_strategy (Union[string,int,float]): Strategy used to determine number of features used at each split 
                 Available options:
-                    "auto": Use the default for lolo (all features)
+                    "auto": Use the default for lolo (all features for regression, sqrt for classification)
                     "log2": Use the base 2 log of the number of features
                     "sqrt": Use the square root of the number of features
                     integer: Set the number of features explicitly
--- a/python/lolopy/learners.py
+++ b/python/lolopy/learners.py
@@ -260,10 +260,13 @@ class RandomForestMixin(BaseLoloLearner):
             use_jackknife (bool): Whether to use jackknife based variance estimates
             bias_learner (BaseLoloLearner): Algorithm used to model bias (default: no model)
             leaf_learner (BaseLoloLearner): Learner used at each leaf of the random forest (default: GuessTheMean)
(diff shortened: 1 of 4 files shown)


python/lolopy/learners.py
```diff
@@ -260,10 +260,13 @@ def __init__(self, num_trees=-1, use_jackknife=True, bias_learner=None,
             use_jackknife (bool): Whether to use jackknife based variance estimates
             bias_learner (BaseLoloLearner): Algorithm used to model bias (default: no model)
             leaf_learner (BaseLoloLearner): Learner used at each leaf of the random forest (default: GuessTheMean)
-            subset_strategy (Union[string,int,float]): Strategy used to determine number of features used at each split in decision
-                trees. Can be "sqrt" for the square root of the number of features, an int to specify the number
-                of features, or a float to set the fraction of the total feature count. "auto" uses "sqrt" for
-                classification problems and 1/3 for regression problems.
+            subset_strategy (Union[string,int,float]): Strategy used to determine number of features used at each split 
+                Available options:
+                    "auto": Use the default for lolo (all features)
```
maxhutch: "all features" is the default for regression, but not classification (which is sqrt, consistent with sklearn).

python/lolopy/loloserver.py
```diff
@@ -60,5 +61,5 @@ def get_java_gateway(reuse=True, skip_devel_version=False):
     if _lolopy_gateway is None or not reuse:
         lolo_path = find_lolo_jar(skip_devel_version)
         _lolopy_gateway = JavaGateway.launch_gateway(classpath=os.path.pathsep.join([
-            os.path.abspath(lolo_path)]), die_on_exit=True)
+            os.path.abspath(lolo_path)]), redirect_stdout=sys.stdout, die_on_exit=True)
```
maxhutch: Does this do anything?

python/lolopy/learners.py
WardLT: Oops, I misread the Scala code

python/lolopy/loloserver.py
WardLT: Yes, it means that anything printed to the standard out from lolo gets printed to the output from lolopy. Otherwise, the output from lolo gets routed to /dev/null. Getting the printouts is useful in debugging (e.g., if I put print statements in lolo). 

Besides debugging, the output has no purpose. But, I decided to keep the printing in lolopy so that I don't have to continually add it when I'm debugging.

maxhutch: Ahh, I didn't know the default was `/dev/null`.  That output is definitely good to have.

PR #164, opened 2019-02-28
Release 2.2.0, 1.0.0: stable API for lolopy and randomized pivots in lolo

maxhutch:
Additions:
 - `randomizePivotLocation` option in tree-based methods (both lolo and lolopy)
 - `LOLOPY_JVM_MEMORY` to control the jvm heap size when the models are large
 - Feature importances in lolopy
 - Constuctor parity for `RandomForest` in lolo and `RandomForestRegressor` in lolopy, including bias and leaf models.
 - Access to lolo metrics in lolopy

Changes:
 - Default feature subset strategy for `RandomForest` regression changed to "all" to match sklearn
 - `RandomForestRegressor` and `RandomForestClassifier` in lolopy have python-style argument names, e.g. `use_jackknife` instead of `useJackknife`


maxhutch:
cc: @WardLT @eantono

PR #166, opened 2019-03-06
Add failing UQ test for multiple candidate prediction

sparadiso:
Making the same prediction (same input vector) batched and one at a time yield different uncertainty estimates.


sparadiso:
Stashing this

PR #167, opened 2019-03-06
Add a failing case between single and multi predict

sparadiso:
There appears to be some inconsistency between single and multi-predict when jack-knife = false.

### DIFF
--- a/src/main/scala/io/citrine/lolo/bags/BaggedResult.scala
+++ b/src/main/scala/io/citrine/lolo/bags/BaggedResult.scala
@@ -66,7 +66,11 @@ case class BaggedSingleResult(
     * @return uncertainty of each prediction
     */
   override def getUncertainty(): Option[Seq[Any]] = Some(Seq(scalarUncertainty))
-  private lazy val scalarUncertainty = Math.sqrt(singleScores.sum) * rescale
+  private lazy val scalarUncertainty = if (useJackknife) {
+    Math.sqrt(singleScores.sum * Math.pow(rescale, 2.0) + Math.pow(bias.getOrElse(0.0), 2.0))
+  } else {
+    bias.getOrElse(0.0)
+  }
 
 
   /**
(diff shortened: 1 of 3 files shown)


src/main/scala/io/citrine/lolo/bags/Bagger.scala
```diff
@@ -120,7 +120,7 @@ case class Bagger(
       val biasTraining = oobErrors.map { case (f, e, u) =>
         // Math.E is only statistically correct.  It should be actualBags / Nib.transpose(i).count(_ == 0)
         // Or, better yet, filter the bags that don't include the training example
-        val bias = Math.max(Math.abs(e) - u, 0)
+        val bias = Math.max(Math.abs(e) - u * ratio, 0)
```
maxhutch: This wasn't contributing the issue, but feels right: since we compute the ratio before training the bias model, we should scale the uncertainty in the bias model's training in the same way that we will rescale it when the model is applied.

src/main/scala/io/citrine/lolo/bags/BaggedResult.scala
```diff
@@ -66,7 +66,11 @@ case class BaggedSingleResult(
     * @return uncertainty of each prediction
     */
   override def getUncertainty(): Option[Seq[Any]] = Some(Seq(scalarUncertainty))
-  private lazy val scalarUncertainty = Math.sqrt(singleScores.sum) * rescale
+  private lazy val scalarUncertainty = if (useJackknife) {
+    Math.sqrt(singleScores.sum * Math.pow(rescale, 2.0) + Math.pow(bias.getOrElse(0.0), 2.0))
```
maxhutch: This is the primary bugfix.  As in `Bagger`, we're applying the scale factor for the jackknife component of the uncertainty only.

maxhutch: @sparadiso @bfolie @latture

maxhutch:
This bug should have affected cases when:
 - There was a bias model but `useJackknife` was false
 - `useJackknife` was false and single predictions where being made of the returned model
 - There was a bias model and single predictions were being made of the returned model.

src/main/scala/io/citrine/lolo/bags/BaggedResult.scala
bfolie: We were actually just discussing a similar change--this brings BaggedSingleResult and BaggedMultiResult into alignment, in that they both follow the same logic for calculating uncertainty. However, there's still the question of why the uncertainty is 0. If useJacknife is false, then the uncertainty is equal to the bias, but the bias learner is predicting 0. Why?

latture: @maxhutch Can we add a test when `rescale != 1.0`? I don't think this that the equation to calculate uncertainty in the multi-result will give the same values in that case.

PR #176, opened 2019-04-02
Uncertainty recalibration

bfolie:
The calculation of the recalibration ratio takes into account points with zero uncertainty, and makes sure the resulting ratio is a real number.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -6,7 +6,7 @@
 
     <groupId>io.citrine</groupId>
     <artifactId>lolo</artifactId>
-    <version>2.2.0</version>
+    <version>2.2.3</version>
     <packaging>jar</packaging>
 
     <name>${project.groupId}:${project.artifactId}</name>
--- a/pom.xml
+++ b/pom.xml
@@ -6,7 +6,7 @@
 
     <groupId>io.citrine</groupId>
     <artifactId>lolo</artifactId>
-    <version>2.2.3</version>
+    <version>2.2.0</version>
     <packaging>jar</packaging>
 
     <name>${project.groupId}:${project.artifactId}</name>
--- a/src/main/scala/io/citrine/lolo/bags/Bagger.scala
+++ b/src/main/scala/io/citrine/lolo/bags/Bagger.scala
@@ -109,7 +109,11 @@ case class Bagger(
     for the training points. If a point has 0 uncertainty, the ratio is 1 if error is also 0, otherwise infinity */
     val ratio = if (uncertaintyCalibration && trainingData.head._2.isInstanceOf[Double] && useJackknife) {
       Async.canStop()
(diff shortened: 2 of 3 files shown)


pom.xml
```diff
@@ -6,7 +6,7 @@
 
     <groupId>io.citrine</groupId>
     <artifactId>lolo</artifactId>
-    <version>2.2.0</version>
+    <version>2.2.3</version>
```
maxhutch: No need to bump versions in `develop`

src/main/scala/io/citrine/lolo/bags/Bagger.scala
```diff
@@ -104,13 +105,18 @@ case class Bagger(
       }
     }
 
-    /* Wrap the models in a BaggedModel object */
+    /* Calculate the uncertainty calibration ratio, which is the 68th percentile of error/uncertainty
+    for the training points. If a point has 0 uncertainty, the ratio is 1 if error is also 0, otherwise infinity */
     val ratio = if (uncertaintyCalibration && trainingData.head._2.isInstanceOf[Double]) {
       Async.canStop()
-      oobErrors.map{case (_, error, uncertainty) => Math.abs(error / uncertainty)}.sorted.drop((oobErrors.size * 0.68).toInt).head
+      oobErrors.map{case (_, error, uncertainty) => uncertainty match {
```
maxhutch: No need to duplicate the `match` here.  Try:
```
      oobErrors.map {
        case (_, 0.0, 0.0) => 1.0
        case (_, _, 0.0) => Double.PositiveInfinity
        case (_, error, uncertainty) => Math.abs(error / uncertainty)
      }.sorted.drop((oobErrors.size * 0.68).toInt).head
```

src/test/scala/io/citrine/lolo/bags/BaggerTest.scala
```diff
@@ -207,6 +209,31 @@ class BaggerTest {
...
+     * then the model will fail to train
+     */
+    val DTLearner = new RegressionTreeLearner(leafLearner = Some(new GuessTheMeanLearner()),
+      numFeatures = 2, randomizePivotLocation = true
+    )
+    val trainedModel: BaggedModel = Bagger(DTLearner,
+      numBags = 16, useJackknife = true, uncertaintyCalibration = true)
+      .train(trainingData)
+      .getModel()
+  }
```
bfolie: I'm not happy with this, because the actual test happens in Bagger.scala, line 120 (where it asserts that the rescaling ratio is real). This test is therefore reliant on external behavior. I'd rather just interrogate the object trainedModel, and make sure that trainedModel.rescale is real. However, there's no direct way to access the rescale field. I could create a getRescale function for the BaggedModel object, but before doing so wanted to check and see if there was a better way to ensure that this model has a real rescale value.

PR #180, opened 2019-04-26
Add failing test where RF fails to train when the response is constant

sparadiso:
This bug was surfaced by Malcolm during the SLAC project.


sparadiso:
LGTM

PR #184, opened 2019-04-30
Resync develop and update the signing key

maxhutch:
This includes and closes #182


maxhutch:
https://search.maven.org/artifact/io.citrine/lolo/2.2.4/jar
https://pypi.org/project/lolopy/1.0.4/

Issue #186, opened 2019-07-01
Bug: `empty head` exception during training


mrupp-citrine:
When training many models via `lolopy` this error sometimes occurs (see below for a full stack trace). It only occurs for small training set sizes (n=10). It occurs non-deterministically.

This error might be related to the Poisson sampling in file `Bagger.scala`, lines 59-69.

Stack trace:
```Traceback (most recent call last):
  File "/Users/mrupp/Data/Citrine/2019-05-01_PredictiveUncertaintyBenchmark/sse-uncertainty-benchmark/smlb/smlb/results.py", line 378, in run
    result = methodf(inputs[traininds], labels[traininds], inputs[validinds])
  File "<string>", line 78, in train_and_predict
  File "/Users/mrupp/Citrine/GitHub/lolo/python/lolopy/learners.py", line 79, in fit
    result = learner.train(train_data, self.gateway.jvm.scala.Some(weights_java))
  File "/Users/mrupp/Local/anaconda3/lib/python3.7/site-packages/py4j/java_gateway.py", line 1286, in __call__
    answer, self.gateway_client, self.target_id, self.name)
  File "/Users/mrupp/Local/anaconda3/lib/python3.7/site-packages/py4j/protocol.py", line 328, in get_return_value
    format(target_id, ".", name), value)
py4j.protocol.Py4JJavaError: An error occurred while calling o52491.train.
: java.lang.UnsupportedOperationException: empty.head
	at scala.collection.immutable.Vector.head(Vector.scala:185)
	at io.citrine.lolo.trees.splits.RegressionSplitter$.getBestSplit(RegressionSplitter.scala:44)
	at io.citrine.lolo.trees.regression.RegressionTreeLearner.train(RegressionTree.scala:70)
	at io.citrine.lolo.trees.regression.RegressionTreeLearner.train(RegressionTree.scala:20)
	at io.citrine.lolo.bags.Bagger.$anonfun$train$10(Bagger.scala:79)
	at io.citrine.lolo.bags.Bagger.$anonfun$train$10$adapted(Bagger.scala:74)
	at scala.collection.parallel.immutable.ParRange$ParRangeIterator.map2combiner(ParRange.scala:104)
	at scala.collection.parallel.ParIterableLike$Map.leaf(ParIterableLike.scala:1052)
	at scala.collection.parallel.Task.$anonfun$tryLeaf$1(Tasks.scala:49)
[...]

PR #188, opened 2019-07-02
Release 2.2.5: bugfix for small training set sizes and large ensembles

maxhutch:
Releasing the fix for https://github.com/CitrineInformatics/lolo/issues/186, which was done in https://github.com/CitrineInformatics/lolo/pull/187.


bfolie:
Many of our stochastically failing tests in mithril seem to involve failed training for small training sets and large ensembles. I'm curious to see if importing this fix from lolo also solves that problem...

PR #189, opened 2019-07-02
Resync develop

maxhutch:
- [x] maven central deploy
- [ ] pypi deploy


maxhutch:
The pypi deployment is broken, but that will take some code to fix.  I captured the issue in #190 and will resolve with a hotfix.

PR #194, opened 2019-12-02
Be less aggresive about rectifying importance scores

maxhutch:
The variance is a sum of importance scores.  Each score is stochastic
and bias corrected, which is noisy and can result in scores that
are negative after the correction (which isn't possible).  We had
been rectifying each of those scores before summing them to get
the variance.

That wasn't a good idea: the whole point of the bias correction
is to shift the estimate down, but our rectifying logic was
in-introducing an upward bias.  Intead, we should rectify _after_
the sum, not before it.  If the result is still negative, then we'll
use the amplitude of the largest (or smallest) importance score
as an estimate of the variance scale.  However, in those cases, the user
should really be using more trees.

### DIFF
--- a/src/main/scala/io/citrine/lolo/bags/BaggedResult.scala
+++ b/src/main/scala/io/citrine/lolo/bags/BaggedResult.scala
@@ -1,9 +1,9 @@
 package io.citrine.lolo.bags
 
-import breeze.linalg.{DenseMatrix, DenseVector, min, norm}
-import breeze.numerics.abs
+import breeze.linalg.{DenseMatrix, DenseVector, norm}
 import io.citrine.lolo.PredictionResult
 import io.citrine.lolo.util.Async
+import org.slf4j.{Logger, LoggerFactory}
 
 /**
   * Interface defining the return value of a [[BaggedModel]]
@@ -356,6 +356,9 @@ class BaggedMultiResult(
 }
 
 object BaggedResult {
+
+  val logger: Logger = LoggerFactory.getLogger(getClass)
+
   /**
    * Make sure the variance is non-negative
    *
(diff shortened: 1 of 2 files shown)


src/main/scala/io/citrine/lolo/bags/BaggedResult.scala
```diff
@@ -306,12 +308,7 @@ class BaggedMultiResult(
       val correction = Math.pow(inverseSize * norm(predMat(::, i) - meanPrediction(i)), 2)
 
       /* The correction is prediction dependent, so we need to operate on vectors */
-      val variancePerRow: DenseVector[Double] = 0.5 * (arg(::, i) - Math.E * correction)
-
-      /* Impose a floor in case any of the variances are negative (hacked to work in breeze) */
```
maxhutch: This is being replaced by https://github.com/CitrineInformatics/lolo/pull/194/files#diff-7c5cef0ff9e4dba1146bc4ff6c4ea36dR368

src/main/scala/io/citrine/lolo/bags/BaggedResult.scala
```diff
@@ -86,13 +91,12 @@ case class BaggedSingleResult(
 
   private lazy val singleScores: Vector[Double] = {
     // Compute the variance of the ensemble of predicted values and divide by the size of the ensemble an extra time
-    val varT = 1.0 / (treePredictions.size * treePredictions.size) * treePredictions.map(y => Math.pow(y - expected, 2.0)).sum
+    val varT = treeVariance / treePredictions.length
```
bfolie: Why divide by ensemble size again?

maxhutch: We're going to subtract the correction from each term in a sum of size `treePredictions.length`, so this division accounts for that subsequent sum.

If we were only computing the variance, then the correction would just be subtracted at the end.  We do it term-by-term to support the `getImportanceScores` and `getInfluenceScores` methods.

src/main/scala/io/citrine/lolo/bags/BaggedResult.scala
mrupp-citrine: Technically, the complement of `rawSum > 0` is not `negative`. What about `rawSum == 0`?

The comment for `rectifyEstimatedVariance` does not explain why `maxEntry` is used in one case and `-scores.min` in the other case.

Why does below part make sense?
```
val floor = Math.abs(scores.min)
if (floor < 0.0)
```

maxhutch: From the docstring:
```
   * If all of the treePredictions are zero, then this will return zero
```
The likelihood that they are non-zero but cancel is infinitesimal.

I'll add to the comments for the others.

PR #195, opened 2019-12-04
Refactor to split BaggedResult impls by type

maxhutch:
The MultiBaggedResult was riddled with if/else logic to handle both
regression and classification in the same class.  This was more confusing
than it was deduplicating, because we don't jackknife for classification.

This splits the classification result out explicitly, making it much
simpler and allowing the regression case to have the correct
T = Double generic class.

### DIFF
--- a/src/main/scala/io/citrine/lolo/bags/BaggedResult.scala
+++ b/src/main/scala/io/citrine/lolo/bags/BaggedResult.scala
@@ -12,7 +12,7 @@ import org.slf4j.{Logger, LoggerFactory}
   * implications.
   */
 trait BaggedResult[+T] extends PredictionResult[T] {
-  protected val predictions: Seq[PredictionResult[T]]
+  def predictions: Seq[PredictionResult[T]]
 
   /**
     * Average the gradients from the models in the ensemble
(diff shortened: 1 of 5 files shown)


maxhutch:
The motivation for this PR is to separate organization and methodological changes (which are upcoming)  and to isolate changes to the learning tasks that they actually affect (regression vs regression and classification).

src/main/scala/io/citrine/lolo/bags/BaggedResult.scala
```diff
@@ -40,25 +40,22 @@ trait BaggedResult extends PredictionResult[Any] {
   * @param predictions for each constituent model
   * @param NibIn       the sample matrix as (N_models x N_training)
   * @param bias        model to use for estimating bias
-  * @param repInput    representative input
   */
 case class BaggedSingleResult(
```
bfolie: Why is `BaggedSingleResult` a case class but the other two are classes?

maxhutch: No good reason.  I got into case classes after the other classes were already written.

PR #198, opened 2019-12-11
Uncertainty decomposition in Regression Interface

maxhutch:
This PR breaks out a specific `RegressionResult` interface and adds both the bias/variance/irreducible error decomposition and the prediction interval / confidence interval distinction.  Further, an `includeNoise` option is added to the existing `getUncertainty` method to distinguish between the prediction and confidence interval understandings of "uncertainty", e.g. "uncertainty in the expectation" vs "uncertainty of the outcome".  Finally, the interface is used for the bagger with ensemble variance serving as a prediction interval and the jackknife estimate as the variance.  The bias and confidence interval may be added pending further study.

### DIFF
--- a/src/main/scala/io/citrine/lolo/PredictionResult.scala
+++ b/src/main/scala/io/citrine/lolo/PredictionResult.scala
@@ -16,7 +16,10 @@ trait PredictionResult[+T] {
   /**
     * Get the uncertainty of the prediction
     *
-    * For example, in regression this is sqrt(bias^2 + variance)
+    * For regression, this should be a standard deviation for the output distribution.
+    * If includeNoise, then it is a prediction interval; if not, then this is a confidence interval
+    *
+    * @param includeNoise whether the uncertainty should account for irreducible noise (i.e. a prediction interval)
     * @return uncertainty of each prediction
     */
   def getUncertainty(includeNoise: Boolean = true): Option[Seq[Any]] = None
@@ -44,12 +47,59 @@ trait PredictionResult[+T] {
   def getGradient(): Option[Seq[Vector[Double]]] = None
 }
 
+/**
+ * Additional regression-specific interface
+ */
 trait RegressionResult extends PredictionResult[Double] {
+  /**
+   * Get the 1-\sigma Prediction Interval for each prediction, if possible
+   *
+   * https://en.wikipedia.org/wiki/Prediction_interval
+   * The prediction interval is typically the sum of the (square) bias, variance, and irreducible error
+   */
   def getPredictionInterval(): Option[Seq[Double]] = None
+
+  /**
+   * Get the 1-\sigma Confidence Interval for each prediction, if possible
+   * https://en.wikipedia.org/wiki/Confidence_interval
+   */
   def getConfidenceInterval(): Option[Seq[Double]] = None
+
+  /**
+   * Get the estimated magnitude of the bias of each prediction, if possible
+   *
+   * See https://en.wikipedia.org/wiki/Bias%E2%80%93variance_tradeoff
+   */
   def getBias(): Option[Seq[Double]] = None
-  def getVariance(): Option[Seq[Double]] = None
+
+  /**
+   * Get the estimated magnitude of the sqrt of the variance of each prediction, if possible
+   *
+   * The sqrt is so that the bias, variance, and irreducible error have the same scale.
+   * See https://en.wikipedia.org/wiki/Bias%E2%80%93variance_tradeoff
+   */
+  def getRootVariance(): Option[Seq[Double]] = None
+
+  /**
+   * Get the estimated magnitude of the irreducible noise of each prediction, if possible
+   *
+   * See https://en.wikipedia.org/wiki/Bias%E2%80%93variance_tradeoff
+   */
   def getIrreducibleError(): Option[Seq[Double]] = None
+
+  /**
+   * Get the "uncertainty", which is the prediction interval if includeNoise and the confidence interval otherwise
+   *
+   * @param includeNoise whether the uncertainty should account for irreducible noise (i.e. a prediction interval)
+   * @return uncertainty of each prediction
+   */
+  override def getUncertainty(includeNoise: Boolean): Option[Seq[Any]] = {
+    if (includeNoise) {
+      getPredictionInterval()
+    } else {
+      getConfidenceInterval()
+    }
+  }
 }
 
 case class MultiResult[T](values: Seq[T]) extends PredictionResult[T] {
--- a/src/main/scala/io/citrine/lolo/PredictionResult.scala
+++ b/src/main/scala/io/citrine/lolo/PredictionResult.scala
@@ -48,6 +48,16 @@ trait PredictionResult[+T] {
 
 /**
  * Additional regression-specific interface
+ * 
+ * This interface is experimental and SHOULD BE REVIEWED before being merged into `master`. 
+ * n particular, an explanation of how the different methods relate to each other, 
+ * how predictive uncertainty is decomposed, and what the assumptions are 
+ * should be added, as these are currently not entirely clear.
+ * 
+ * For example, does the interface assume that the predictions are the mean of a predictive distribution 
+ * (as opposed to, for example, the median, or the value with highest probability)? 
+ * Does it assume the predictive distribution to be normal? 
+ * Such assumptions are fine, but should be explicitly stated.
  */
 trait RegressionResult extends PredictionResult[Double] {
   /**
--- a/src/main/scala/io/citrine/lolo/PredictionResult.scala
+++ b/src/main/scala/io/citrine/lolo/PredictionResult.scala
@@ -44,6 +44,14 @@ trait PredictionResult[+T] {
   def getGradient(): Option[Seq[Vector[Double]]] = None
 }
 
+trait RegressionResult extends PredictionResult[Double] {
+  def getPredictionInterval(): Option[Seq[Double]] = None
+  def getConfidenceInterval(): Option[Seq[Double]] = None
+  def getBias(): Option[Seq[Double]] = None
+  def getVariance(): Option[Seq[Double]] = None
+  def getIrreducibleError(): Option[Seq[Double]] = None
+}
+
 case class MultiResult[T](values: Seq[T]) extends PredictionResult[T] {
   /**
     * Get the expected values for this prediction
--- a/src/main/scala/io/citrine/lolo/PredictionResult.scala
+++ b/src/main/scala/io/citrine/lolo/PredictionResult.scala
@@ -56,7 +56,7 @@ trait RegressionResult extends PredictionResult[Double] {
    *
    * Observations of the predicted variable are expected to have a stddev that matches this value.
    * This statistic is related to the https://en.wikipedia.org/wiki/Prediction_interval
-   * It does not include estimated bias, if applicable
+   * It does not include estimated bias, even if the regression result contains a bias estimate.
    */
   def getStdDevObs(): Option[Seq[Double]] = None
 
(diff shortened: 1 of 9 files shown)


maxhutch:
@mrupp-citrine @sparadiso @bfolie @latture : I probably want to refine the user-facing documentation before merging, but I'd like some early feedback on this design.

cc: @WardLT

src/main/scala/io/citrine/lolo/bags/BaggedResult.scala
```diff
@@ -60,21 +59,14 @@ case class BaggedSingleResult(
...
-    *
-    * @return uncertainty of each prediction
-    */
-  override def getUncertainty(): Option[Seq[Any]] = Some(Seq(scalarUncertainty))
+  override def getPredictionInterval(): Option[Seq[Double]] = Some(Seq(treeVariance))
+
+  override def getRootVariance(): Option[Seq[Double]] = Some(Seq(scalarUncertainty))
 
   private lazy val scalarUncertainty: Double = {
-    if (useJackknife) {
```
bfolie: We always use Jackknife now?

maxhutch: We'll always use it to get the variance component of the error, yes.  We may want to preserve a way to not carry the `Nib` matrix around if we're never going to use it, but the laziness of this declaration prevents us from paying the runtime cost and I'm not particularly concerned about memory right now.

maxhutch:
Note: the failing tests are because the interface changed but the python binding wasn't updated.  If we like this design, I can update the python binding.

src/main/scala/io/citrine/lolo/PredictionResult.scala
```diff
@@ -44,6 +47,61 @@ trait PredictionResult[+T] {
...
+   */
+  def getIrreducibleError(): Option[Seq[Double]] = None
+
+  /**
+   * Get the "uncertainty", which is the prediction interval if includeNoise and the confidence interval otherwise
+   *
+   * @param includeNoise whether the uncertainty should account for irreducible noise (i.e. a prediction interval)
+   * @return uncertainty of each prediction
+   */
+  override def getUncertainty(includeNoise: Boolean): Option[Seq[Any]] = {
```
bfolie: Thoughts on giving `includeNoise` a default value, in order to limit changes to the interface? My inclination is that we should use a default if it won't be confusing, and that people would generally assume the uncertainty includes irreducible noise.

maxhutch: Totally; I meant to do this everywhere (default true) and just missed this one.

mrupp-citrine: **`getUncertainty`**

This seems a reasonable interface given the  constraint of backwards compatibility. 

**`RegressionResult`**

* how extensible is this interface with respect to heteroskedastic, and potentially even non-additive noise?
* the descriptions mention "1-sigma", but the distributions are not specified. Is normality assumed here?
* related, the interface crucially relies on intervals; while this is probably fine, it does preclude extensions toward other distributions (for example, multi-modal ones) 
* `getIrreducibleError` is not wrong, but what has motivated this choice over alternatives based on measurement noise, or label noise?

klennox-citrine: `getUncertainty`
Would it be possible to use a 'type' argument instead of the 'includeNoise' boolean? The meaning of includeNoise wouldn't be intuitive to an outside user, while something like 'NormConfIntWidth95' or 'SamplingSDMean' would. This would also allow for extensibility later if we wanted to include other forms of variability measurements. It would also be backwards compatible with a default value of, for example, 'SamplingSDMean' or 'SamplingSDReg'.

`getPredictionInterval` `getConfidenceInterval`
Prediction and confidence interval calculators usually include an argument for setting the level of the interval (e.g. 0.68, 0.95). Allowing this to vary is pretty easy when assuming normality, since you just use the normal or t quantile function to get the right multiplier for the SD. Generally, level=0.68 will not be assumed when you say prediction or confidence interval. Another option would be to stick with 1SD, but rename the methods after the SD being estimated. (This may be tricky for a prediction interval. Off the top of my head, I'm not sure what the base interval width should be called. Prediction error maybe?) 

`getRootVariance`
Why not use getStandardDeviation?

`getIrreducibleError`
Agree that another name might be more interpretable to the wider community. getObservationSD? (Measurement error is also a possibility, but has a different and very specific meaning in stats that we may want to make use of later.

maxhutch: > **`RegressionResult`**
> 
> * how extensible is this interface with respect to heteroskedastic, and potentially even non-additive noise?

This assumes heteroskedastic additive noise.  Homoskedastic is a special case of that.  If you wanted to express "I know my noise is homoskedastic" you'd do that in the `Learner` not in the `PredictionResult`

> * the descriptions mention "1-sigma", but the distributions are not specified. Is normality assumed here?
> * related, the interface crucially relies on intervals; while this is probably fine, it does preclude extensions toward other distributions (for example, multi-modal ones)

Returning generic distributions is beyond the scope.  "1-sigma" was meant as a "give me the 2nd moment" in a series-expansion kind of way.

> * `getIrreducibleError` is not wrong, but what has motivated this choice over alternatives based on measurement noise, or label noise?

Alternative choices of names?  Lack of creativity.  I'm very open to suggestions.

maxhutch: > `getUncertainty`
> Would it be possible to use a 'type' argument instead of the 'includeNoise' boolean? The meaning of includeNoise wouldn't be intuitive to an outside user, while something like 'NormConfIntWidth95' or 'SamplingSDMean' would. This would also allow for extensibility later if we wanted to include other forms of variability measurements. It would also be backwards compatible with a default value of, for example, 'SamplingSDMean' or 'SamplingSDReg'.
> 

What would be the list of options (and what do they mean)?  Are they sufficiently general to exist in the interface with multiple implementing methods in mind?  I'm not quite sure what `SamplingSDMean` would be, for example.

> `getPredictionInterval` `getConfidenceInterval`
> Prediction and confidence interval calculators usually include an argument for setting the level of the interval (e.g. 0.68, 0.95). Allowing this to vary is pretty easy when assuming normality, since you just use the normal or t quantile function to get the right multiplier for the SD. Generally, level=0.68 will not be assumed when you say prediction or confidence interval. Another option would be to stick with 1SD, but rename the methods after the SD being estimated. (This may be tricky for a prediction interval. Off the top of my head, I'm not sure what the base interval width should be called. Prediction error maybe?)

I like this idea, but note that it contrasts with the standard moment expansion approach that I suggested in my response to @mrupp-citrine .
 
> `getRootVariance`
> Why not use getStandardDeviation?

To distinguish between sqrt(variance) and sqrt(variance + bias^2).  Given the term "bias-variance tradeoff", its not clear to me in what scope "RootVariance" and "StandardDeviation" are synonymous.

> `getIrreducibleError`
> Agree that another name might be more interpretable to the wider community. getObservationSD? (Measurement error is also a possibility, but has a different and very specific meaning in stats that we may want to make use of later.

`getObservationSD` seems reasonable.  What about `getNoise`?

mrupp-citrine:
## Some remarks

### Bias estimates

It is currently unclear whether we should report bias estimates or immediately correct for them.
Preliminary @maxhutch results indicate that substantial (ca. 30%) improvement in prediction errors (RMSE) could result from correction. Besides `getBias`, this affects other routines (e.g., `getStdDevObs`); it therefore complicates the interface, with unclear cost-benefit trade-off. 

I would argue, conditional on further confirmatory results by Hutch, to immediately correct for estimated bias, and to not expose it in the interface. Improved prediction errors are the most direct benefit for us, and the interface complication does not seem worth knowing the bias. It is also unclear if and how other predictors would support it.

### Error decomposition

The, or at least a, accepted decomposition of predictive uncertainties in the literature seems to be into model uncertainty ('epistemic') and signal uncertainty ('aleatoric'). This seems reasonable. Another alternative would be to distinguish contributions to signal uncertainty (instrument, materials sample, user error, ...). However, while this would be related to our detailed data model, we have no way to distinguish those contributions. Reporting only total error estimate seems overly simple; having both model and signal uncertainty is useful for sequential learning. For bias estimate see above.

### Reported quantities

I feel that we should keep the reported quantities to a minimum unless there is clear evidence that we need more. The minimal set seems to me the model and signal uncertainties.

### Naming

The current naming would simplify if we correct for the estimated bias instead of reporting it. The bias estimate issue should be resolved before naming is discussed further.

## The discussion so far

Mostly for my benefit, but including it for reference.

`getUncertainty`

Has an `includeNoise` parameter to distinguish epistemic from aleatoric uncertainty estimates. For backwards compatibility, this is `true` by default.

Alternatively, a 'type' argument could be provided, for example, with values `NormConfIntWidth95`, `SamplingSDMean`, and others
* more flexible and extensible (to other variability measures)
* less simple, but perhaps more explicit, interface
* unclear which values should be supported
* other methods would have to support this; how to do that while limiting implementation effort?

`RegressionResult` interface

* the interface covers heteroskedastic noise, which is important.
* it is limited to normal distribution, which is a reasonable compromise. (*)
* it is unclear whether normality is assumed /by the interface/

(*) distributions could be represented via samples (empirical cumulative distribution function), which seems related to sampling in our graphical modeling approach?

`getPredictionInterval`, `getConfidenceInterval`
* a parameter (with default value one standard deviation) could be provided to allow for different quantile levels
* conflicts with moment-expansion interpretation

`getRootVariance`
* why not use getStandardDeviation? to distinguish between sqrt(variance) and sqrt(variance + bias^2)

`getIrreducibleError`

* Naming is unclear. Suggestions: `getObservationSD`, `getMeasurementError` (too specific), `getNoise`

maxhutch:
Re `Bias estimates` from @mrupp-citrine : he and I spoke and agreed that we're not yet sure if we always want to correct the bias.  For example, when doing active learning, the bias contribution to the uncertainty could help steer the experimental designs better than the variance alone.  We just need to test it, and for that we need to get it into a development interface (since active learning is beyond the scope of lolo).

I'm going to look for some way to flag the methods as "experimental" to signal to any users that they are more likely to change in the future.

src/main/scala/io/citrine/lolo/PredictionResult.scala
```diff
@@ -120,14 +119,16 @@ trait RegressionResult extends PredictionResult[Double] {
   def getQuantileMean(quantile: Double): Option[Seq[Double]] = None
 
   /**
-   * Get the estimated magnitude of the bias of each prediction, if possible
+   * **EXPERIMENTAL** Get the estimated magnitude of the bias of each prediction, if possible
```
mrupp-citrine: "magnitude" sounds like an absolute value (only the magnitude, not the sign), but the comment below says it's signed.

src/main/scala/io/citrine/lolo/PredictionResult.scala
```diff
@@ -44,6 +46,112 @@ trait PredictionResult[+T] {
   def getGradient(): Option[Seq[Vector[Double]]] = None
 }
 
+/**
+ * Additional regression-specific interface
+ */
```
mrupp-citrine: Suggest to add: "This interface is experimental and SHOULD BE REVIEWED before being merged into `master`. In particular, an explanation of how the different methods relate to each other, how predictive uncertainty is decomposed, and what the assumptions (*) are should be added, as these are currently not entirely clear.
(*) For example, does the interface assume that the predictions are the mean of a predictive distribution (as opposed to, for example, the median, or the value with highest probability)? Does it assume the predictive distribution to be normal? Such assumptions are fine, but should be explicitly stated."

src/main/scala/io/citrine/lolo/PredictionResult.scala
```diff
@@ -49,55 +48,116 @@ trait PredictionResult[+T] {
 
 /**
  * Additional regression-specific interface
+ *
+ * This interface is experimental and SHOULD BE REVIEWED before being merged into `master`.
+ * n particular, an explanation of how the different methods relate to each other,
```
bfolie:
```suggestion
 * In particular, an explanation of how the different methods relate to each other,
```

PR #199, opened 2020-01-09
Support random rotations of features

gregor-robinson:
This PR introduces a transformer that rotates input feature space before passing the transformed data to a base learner, with a specific aim of applying random rotations. There is reason to believe that random rotations of feature space can improve the performance of various kinds of learners --- such as by increasing entropy across an ensemble, or by decreasing the incidence of unimportant feature subsets for methods that use ensembles over random feature subsets. One promising application, random rotation ensembles introduced by Blaser & Fryzlewicz, trains each base learner in an ensemble method on a different random rotation of feature space; predictions are performed using the same rotations that were applied to the training data. Blaser & Fryzlewicz (2016) found large improvement in RMSE using this approach on Fisher's iris dataset.

Note: there are a few tests indicated by TODOs that I still want to write before merging this, but I would appreciate eyes on the implementation for feedback. Some questions for the team: What's an appropriate way to handle None in `getGradient`? Are there any other cases you want to see in the unit tests?

### DIFF
--- a/src/main/scala/io/citrine/lolo/transformers/FeatureRotator.scala
+++ b/src/main/scala/io/citrine/lolo/transformers/FeatureRotator.scala
@@ -93,7 +93,7 @@ class RotatedFeatureTrainingResult(
   * @param baseModel model to which to delegate prediction on rotated features
   * @param rotatedFeatures indices of features to rotate
   * @param trans matrix to apply to features
-  * @tparam T
+  * @tparam T label type
   */
 class RotatedFeatureModel[T](
                              baseModel: Model[PredictionResult[T]],
@@ -179,15 +179,11 @@ object FeatureRotator {
     * @return list of feature indices that are doubles
     */
   def getDoubleFeatures(rep: Vector[Any]): IndexedSeq[Int] = {
-    rep.indices.collect( i =>
-      rep(i) match {
-        case _: Double => i
-      }
-    )
+    rep.indices.filter(i => rep(i).isInstanceOf[Double])
   }
 
   /**
-   * Apply rotation to a vectors.
+   * Apply rotation to a vector.
    *
    * @param input vector to rotate
    * @param featuresToRotate vector of feature indices included in rotation
--- a/src/main/scala/io/citrine/lolo/transformers/FeatureRotator.scala
+++ b/src/main/scala/io/citrine/lolo/transformers/FeatureRotator.scala
@@ -1,0 +1,225 @@
+package io.citrine.lolo.transformers
+
+import io.citrine.lolo._
+import breeze.linalg.{DenseMatrix, DenseVector, diag, qr}
+import breeze.linalg.qr.QR
+import breeze.numerics.signum
+import breeze.stats.distributions.Gaussian
+
+/**
+  * Rotate the training data before passing along to a base learner
+  *
+  * This may be useful for improving randomization in random forests,
+  * especially when using random feature selection without bagging.
+  *
+  * Created by gregor-robinson on 2020-01-02.
+  */
+case class FeatureRotator(baseLearner: Learner) extends Learner {
+
+  /**
+    * Create linear transformations for continuous features and labels & pass data through to learner
+    *
+    * @param trainingData to train on
+    * @param weights      for the training rows, if applicable
+    * @return training result containing a model
+    */
+  override def train(
+                     trainingData: Seq[(Vector[Any], Any)],
+                     weights: Option[Seq[Double]]
+                    ): RotatedFeatureTrainingResult = {
+    val featuresToRotate = FeatureRotator.getDoubleFeatures(trainingData.head._1)
+    val trans = FeatureRotator.getRandomRotation(featuresToRotate.length)
+
+    val (inputs, labels) = trainingData.unzip
+    val rotatedTrainingData = FeatureRotator.applyRotation(inputs, featuresToRotate, trans).zip(labels)
+    val baseTrainingResult = baseLearner.train(rotatedTrainingData, weights)
+
+    new RotatedFeatureTrainingResult(baseTrainingResult, featuresToRotate, trans)
+  }
+}
+
+class MultiTaskFeatureRotator(baseLearner: MultiTaskLearner) extends MultiTaskLearner {
+
+  /**
+    * Create linear transformations for continuous features and labels & pass data through to learner
+    *
+    * @param inputs  to train on
+    * @param labels  sequence of sequences of labels
+    * @param weights for the training rows, if applicable
+    * @return a sequence of training results, one for each label
+    */
+  override def train(
+                     inputs: Seq[Vector[Any]],
+                     labels: Seq[Seq[Any]],
+                     weights: Option[Seq[Double]]
+                    ): Seq[RotatedFeatureTrainingResult] = {
+    val featuresToRotate = FeatureRotator.getDoubleFeatures(inputs.head)
+    val trans = FeatureRotator.getRandomRotation(inputs.head.length)
+    val rotatedFeatures = FeatureRotator.applyRotation(inputs, featuresToRotate, trans)
+    val baseTrainingResult = baseLearner.train(rotatedFeatures, labels, weights)
+
+    baseTrainingResult.map { case (base) =>
+      new RotatedFeatureTrainingResult(base, featuresToRotate, trans)
+    }
+  }
+}
+
+/**
+  * Training result bundling the base learner's training result with the list of rotated features and the transformation
+  *
+  * @param baseTrainingResult training result to which to delegate prediction on rotated features
+  * @param rotatedFeatures indices of features to rotate
+  * @param trans matrix to apply to features
+  */
+class RotatedFeatureTrainingResult(
+                                   baseTrainingResult: TrainingResult,
+                                   rotatedFeatures: IndexedSeq[Int],
+                                   trans: DenseMatrix[Double]
+                                  ) extends TrainingResult {
+
+  /**
+    * Get the model contained in the training result
+    *
+    * @return the model
+    */
+  override def getModel(): Model[PredictionResult[Any]] = {
+    new RotatedFeatureModel(baseTrainingResult.getModel(), rotatedFeatures, trans)
+  }
+}
+
+/**
+  * Model bundling the base learner's model with the list of rotated features and the transformation
+  *
+  * @param baseModel model to which to delegate prediction on rotated features
+  * @param rotatedFeatures indices of features to rotate
+  * @param trans matrix to apply to features
+  * @tparam T
+  */
+class RotatedFeatureModel[T](
+                             baseModel: Model[PredictionResult[T]],
+                             rotatedFeatures: IndexedSeq[Int],
+                             trans: DenseMatrix[Double]
+                            ) extends Model[PredictionResult[T]] {
+
+  /**
+    * Transform the inputs and then apply the base model
+    *
+    * @param inputs to apply the model to
+    * @return a RotatedFeaturePredictionResult which includes, at least, the expected outputs
+    */
+  override def transform(inputs: Seq[Vector[Any]]): RotatedFeaturePrediction[T] = {
+    val rotatedInputs = FeatureRotator.applyRotation(inputs, rotatedFeatures,  trans)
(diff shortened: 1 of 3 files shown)


src/main/scala/io/citrine/lolo/transformers/FeatureRotator.scala
```diff
@@ -0,0 +1,225 @@
...
+  override def getUncertainty(): Option[Seq[Any]] = baseResult.getUncertainty()
+
+  /**
+    * Get the gradient or sensitivity of each prediction
+    *
+    * @return a vector of doubles for each prediction
+    */
+  override def getGradient(): Option[Seq[Vector[Double]]] = {
+    var baseGradient = baseResult.getGradient()
+    // TODO: how should this handle None gradients?
```
maxhutch: Then it should just return none (which is the behavior below, I believe).

src/main/scala/io/citrine/lolo/transformers/FeatureRotator.scala
```diff
@@ -0,0 +1,225 @@
...
+  /**
+    * Get list of feature indices that make sense to rotate.
+    *
+    * @param rep representative vector of features
+    * @return list of feature indices that are doubles
+    */
+  def getDoubleFeatures(rep: Vector[Any]): IndexedSeq[Int] = {
+    rep.indices.collect( i =>
+      rep(i) match {
+        case x if x.isInstanceOf[Double] => i
```
maxhutch: Does `case _: Double => i` not work because of type erasure?

I'd be a bit surprised if this collect syntax works.  Can you test it when there are some categorical inputs?  You can use [this pattern](https://github.com/CitrineInformatics/lolo/blob/develop/src/test/scala/io/citrine/lolo/learners/RandomForestTest.scala#L18) to generate mixed real/categorical input vectors.

gregor-robinson: This was an experiment that I neglected to roll back, trying to fix IntelliJ's complaint that a `PartialFunction[Int,NotInferredB]` is required. I now suspect that complaint reflects a static analysis bug, but I'd appreciate if you could check this spot for potential issues the static analyzer is correctly flagging.

gregor-robinson:
I've finished and pushed the tests I had in mind.

src/main/scala/io/citrine/lolo/transformers/FeatureRotator.scala
bfolie: `rep.indices.collect` allows you to collect the elements of `rep.indices` that meet some criterion, but `rep.indices` is just a list of indices (0, 1, 2, 3, ...) and even though that's what you're collecting, you're filtering on something else (the type of `rep(i)`).
I think the simplest way to do this is 
```rep.indices.filter(i => rep(i).isInstanceOf[Double])```
If you want to avoid `isInstanceOf`, then you could do
```
rep.zip(rep.indices).collect {
      case (_: Double, i: Int) => i
    }
```

File: src/main/scala/io/citrine/lolo/transformers/FeatureRotator.scala [hunk 3]
```diff
@@ -0,0 +1,225 @@
...
+  def getDoubleFeatures(rep: Vector[Any]): IndexedSeq[Int] = {
+    rep.indices.collect( i =>
+      rep(i) match {
+        case _: Double => i
+      }
+    )
+  }
+
+  /**
+   * Apply rotation to a vectors.
```
bfolie:
```suggestion
   * Apply rotation to a vector.
```

src/main/scala/io/citrine/lolo/transformers/FeatureRotator.scala
```diff
@@ -0,0 +1,225 @@
...
+  }
+}
+
+/**
+  * Model bundling the base learner's model with the list of rotated features and the transformation
+  *
+  * @param baseModel model to which to delegate prediction on rotated features
+  * @param rotatedFeatures indices of features to rotate
+  * @param trans matrix to apply to features
+  * @tparam T
```
bfolie: What is T?

src/main/scala/io/citrine/lolo/transformers/FeatureRotator.scala
gregor-robinson: I do prefer that. Thanks!

src/main/scala/io/citrine/lolo/transformers/FeatureRotator.scala
```diff
@@ -0,0 +1,221 @@
...
+  *
+  * @param baseTrainingResult training result to which to delegate prediction on rotated features
+  * @param rotatedFeatures indices of features to rotate
+  * @param trans matrix to apply to features
+  */
+class RotatedFeatureTrainingResult(
+                                   baseTrainingResult: TrainingResult,
+                                   rotatedFeatures: IndexedSeq[Int],
+                                   trans: DenseMatrix[Double]
+                                  ) extends TrainingResult {
```
maxhutch: Can you implement `getLoss` or `getPredictedVsActual` for any cases?

src/main/scala/io/citrine/lolo/transformers/FeatureRotator.scala
```diff
@@ -0,0 +1,221 @@
...
+}
+
+/**
+  * Training result bundling the base learner's training result with the list of rotated features and the transformation
+  *
+  * @param baseTrainingResult training result to which to delegate prediction on rotated features
+  * @param rotatedFeatures indices of features to rotate
+  * @param trans matrix to apply to features
+  */
+class RotatedFeatureTrainingResult(
```
maxhutch: Why not a case class here too?

src/main/scala/io/citrine/lolo/transformers/FeatureRotator.scala
```diff
@@ -0,0 +1,221 @@
...
+
+    val (inputs, labels) = trainingData.unzip
+    val rotatedTrainingData = FeatureRotator.applyRotation(inputs, featuresToRotate, trans).zip(labels)
+    val baseTrainingResult = baseLearner.train(rotatedTrainingData, weights)
+
+    new RotatedFeatureTrainingResult(baseTrainingResult, featuresToRotate, trans)
+  }
+}
+
+class MultiTaskFeatureRotator(baseLearner: MultiTaskLearner) extends MultiTaskLearner {
```
maxhutch: Why not a case class here too?

src/main/scala/io/citrine/lolo/transformers/FeatureRotator.scala
```diff
@@ -0,0 +1,221 @@
...
+  }
+
+  /**
+    * Get list of feature indices that make sense to rotate.
+    *
+    * @param rep representative vector of features
+    * @return list of feature indices that are doubles
+    */
+  def getDoubleFeatures(rep: Vector[Any]): IndexedSeq[Int] = {
+    rep.indices.filter(i => rep(i).isInstanceOf[Double])
```
maxhutch: This is fine since `rep: Vector` has O(1) indexing, but know that you could use `.zipWithIndex.filter{case (i, r) => r.isInstanceOf[Double]}.map(_._1)` to stream it if it were just `rep: Seq`.  I'm never sure a priori what will or won't work with erasure, but you might be able to do `.zipWithIndex.collect{case (i, r: double) => i}` or at least `.zipWithIndex.collect{case (i, r) if r.isInstanceOf[Double] => i}` as well.

src/test/scala/io/citrine/lolo/transformers/FeatureRotatorTest.scala
```diff
@@ -0,0 +1,227 @@
...
+    val rotatedModel = rotatedLearner.train(data).getModel()
+    val rotatedResult = rotatedModel.transform(data.map(_._1)).getExpected()
+
+    result.zip(rotatedResult).foreach { case (free: Double, rotated: Double) =>
+      assert(Math.abs(free - rotated) < 1.0e-9, s"${free} and ${rotated} should be the same")
+    }
+  }
+
+  /**
+   * Regression trees should not depend on rotation
```
maxhutch: Regression trees _should_ depend on the rotation, but they should memorize their inputs regardless (if grown to full depth).  Could you add a check that non-training inputs do change?

src/test/scala/io/citrine/lolo/transformers/FeatureRotatorTest.scala
```diff
@@ -0,0 +1,227 @@
...
+    val rotatedModel = rotatedLearner.train(data).getModel()
+    val rotatedResult = rotatedModel.transform(data.map(_._1)).getExpected()
+
+    result.zip(rotatedResult).foreach { case (free: Double, rotated: Double) =>
+      assert(Math.abs(free - rotated) < 1.0e-9, s"${free} and ${rotated} should be the same")
+    }
+  }
+
+  /**
+   * Classification trees should not depend on rotation
```
maxhutch: Likewise for regression trees.

src/test/scala/io/citrine/lolo/transformers/FeatureRotatorTest.scala
```diff
@@ -0,0 +1,227 @@
...
+    val rotatedModel = rotatedLearner.train(classificationData).getModel()
+    val rotatedResult = rotatedModel.transform(classificationData.map(_._1)).getExpected()
+
+    result.zip(rotatedResult).foreach { case (free: Any, rotated: Any) =>
+      assert(free == rotated, s"${free} and ${rotated} should be the same")
+    }
+  }
+
+  /**
+   * Test that multitask rotation has the proper performance and invariants
```
maxhutch: Likewise here as well.

src/main/scala/io/citrine/lolo/transformers/FeatureRotator.scala
gregor-robinson: Thanks, so many options! I'll leave it as is, unless you have strong preference toward another variant.

gregor-robinson:
Thanks @maxhutch! The invariance for training data was by design, but I agree it's important to test non-training data. To that end, I have added tests that verify that the rotated feature models predict incorrect values when passed rotated data (i.e. the data are rotated twice). This has the advantage of letting me reuse those test data for another new test: verify that the base learner acting on the rotated data reproduce the training labels.

I have implemented pass-through methods and added tests for them, and changed all classes to case classes.

src/main/scala/io/citrine/lolo/transformers/FeatureRotator.scala
```diff
@@ -85,6 +85,10 @@ class RotatedFeatureTrainingResult(
   override def getModel(): Model[PredictionResult[Any]] = {
     new RotatedFeatureModel(baseTrainingResult.getModel(), rotatedFeatures, trans)
   }
+
+  override def getLoss(): Option[Double] = baseTrainingResult.getLoss()
+
+  override def getPredictedVsActual(): Option[Seq[(Vector[Any], Any, Any)]] = baseTrainingResult.getPredictedVsActual()
```
maxhutch: I think you need to un-rotate the feature vectors here.

src/test/scala/io/citrine/lolo/transformers/FeatureRotatorTest.scala
```diff
@@ -80,6 +80,22 @@ class FeatureRotatorTest {
     }
   }
 
+  /**
+    * Functions getLoss and getPredictedVsActual should pass through to base learner.
+    */
+  @Test
+  def testPassthroughFunctions(): Unit = {
+    val rotatedTrainingResult = FeatureRotator(GuessTheMeanLearner()).train(data)
```
maxhutch: `GuessTheMeanLearner` is probably too trivial of a case, here, since the predicted labels will be uniform.  I'd recommend a single decision tree (since its still fast).

gregor-robinson: Fair enough. Done.

src/main/scala/io/citrine/lolo/transformers/FeatureRotator.scala
gregor-robinson: You're right. Fixed.

PR #200, opened 2020-02-03
Expose random rotations in python interface.

gregor-robinson:


src/main/scala/io/citrine/lolo/learners/RandomForest.scala
```diff
@@ -67,7 +69,7 @@ case class RandomForest(
           maxDepth = maxDepth,
           splitter = RegressionSplitter(randomizePivotLocation)
         )
-        val bagger = Bagger(DTLearner,
+        val bagger = Bagger(if (randomlyRotateFeatures) FeatureRotator(DTLearner) else DTLearner,
```
maxhutch: So nice...

PR #201, opened 2020-02-26
Add Shapley prototype.

gregor-robinson:
This is a first scratch at implementing Shapley feature importance in Lolo. A few open questions / problems:
* ~Why does the last set of assertions in testShapley pass? It's consistent with the reference implementation, but inconsistent with my pen-and-paper. Most likely I'm making a conceptual mistake in my reading of the paper, but it seems important to rule out the possibility that there is a bug in the original pseudocode that is reproduced in the reference implementation. There is, after all, a different bug in the paper (but it's not reproduced in the reference implementation).~ I was incorrectly counting entries in the cold branch.
* How should we make this play nicely across different label types?
* ~Specifically, how should this work for a classifier. It's easy to imagine computing Shap on classification probability vectors... does the intuition and utility still hold?~ Yes, it works perfectly for class probabilities.
* ~Once we settle on an API at the tree level, it will be straightforward to extend this to the forest level.~
* How to handle leaf learners other than GuessTheMeanLearner?
* There are some optimizations that can be made without much effort, such as reducing copying overhead. Benchmarks on relevant problem sizes can identify whether/when this would a good use of engineering time. `UPDATE`: this would be worth doing for data more sizable than a few hundred, around which time the computation starts taking minutes. Basic instrumentation suggests that the copying overhead is indeed the leading source of overhead, even without considering the (probably second-order) effect of heap-thrashing due to excess allocation/GC.


src/main/scala/io/citrine/lolo/trees/FeaturePath.scala
```diff
@@ -0,0 +1,96 @@
...
+/**
+  * Description of a feature's role in computing TreeSHAP
+  *
+  * @param featureIndex index of the feature this node describes.
+  * @param zeroFraction fraction of paths flowing through this node with this feature excluded.
+  * @param oneFraction fraction of one paths flowing through this node with this feature included.
+  * @param pathWeight summation weight for this path, taking into account combinatorial factor.
+  */
+case class FeatureNode(
+                  var featureIndex: Int,
```
maxhutch: `var` in case classes is a big anti-pattern.  Case classes should almost always be immutable.  This mutability would be very confusing to future developers, possibly leading to bugs.

gregor-robinson: Thanks. I've switched to regular classes where the algorithm uses mutability. It would be possible to refactor to use immutable case classes, but that would add complication and I'd like to get a MVP shipped.

gregor-robinson:
This is ready for another round of architectural feedback.

Reworked to live in ModelNode, toward wrapping in Bagger.

Switched to a vector-valued return type to provide a path toward generalizing to classification and multi-objective. I'm not particularly pleased with how this feels... it seems a little clunky to return a bunch of length-1 DenseVectors in the case of simple regression, but other options coming to mind seem like even more overkill (such as adding a ShapleyResult abstract class).

src/main/scala/io/citrine/lolo/trees/Nodes.scala
```diff
@@ -43,12 +79,18 @@ trait ModelNode[T <: PredictionResult[Any]] extends Serializable {
   * @param split to decide which branch to take
   * @param left  branch node
   * @param right branch node
+  * @param numFeatures number of features, used for Shapley computation
+  * @param outputDimension dimension of model output, used for Shapley computation
```
maxhutch: Its not clear to me what this number is.  Can you give an example, like "1 for single-task regression" or "equal to the number of categories in classification"?

gregor-robinson:
@maxhutch This is ready for review to merge into develop.

> Could you add a description of the algorithm (or a link to a description of the algorithm) that you're using somewhere. Also the output (again, link is fine).

I added a [link to the algorithm](https://arxiv.org/abs/1802.03888) in commit `85a2d0d`, but it's a `doozy` (in @bfolie's apt words). Intuition about the output is provided in that reference as well.

src/test/scala/io/citrine/lolo/bags/BaggerTest.scala
```diff
@@ -360,17 +360,29 @@ class BaggerTest {
...
+      // Do a quick sanity check on the output format.
       assert(shapley.length == nCols)
       assert(shapley.forall {
         _.length == 1
       })
+
+      // Compute the mean shap value over trees and ensure the bagged model gives the same result.
+      val treeMean = trees.map { t =>
+        t.shapley(x).get.map { _(0) / trees.length }
+      }.toVector.transpose.map { case ts: Seq[Double] => ts.reduce[Double] { case (y1: Double, y2: Double) => y1 + y2 } }
```
bfolie, line 381: Why not `ts.sum`?

src/test/scala/io/citrine/lolo/bags/BaggerTest.scala
```diff
@@ -360,17 +360,29 @@ class BaggerTest {
...
+
+      // Do a quick sanity check on the output format.
       assert(shapley.length == nCols)
       assert(shapley.forall {
         _.length == 1
       })
+
+      // Compute the mean shap value over trees and ensure the bagged model gives the same result.
+      val treeMean = trees.map { t =>
+        t.shapley(x).get.map { _(0) / trees.length }
```
bfolie, line 380: I think you can use `flatten` here

src/main/scala/io/citrine/lolo/Model.scala
```diff
@@ -9,7 +11,21 @@ trait Model[+T <: PredictionResult[Any]] extends Serializable {
...
+    *
+    * @param input for which to compute feature attributions.
+    * @param omitFeatures feature indices to omit in computing Shapley values
+    * @return array of vector-valued attributions for each feature
+    *         One DenseVector[Double] per feature, each of length equal to the output dimension.
+    */
+  def shapley(
+               input: Vector[Any],
+               omitFeatures: Set[Int] = Set()
+             ): Option[Vector[DenseVector[Double]]] = None
```
bfolie, line 29: Why specify the internal vector as dense?

src/main/scala/io/citrine/lolo/Model.scala
```diff
@@ -9,7 +11,21 @@ trait Model[+T <: PredictionResult[Any]] extends Serializable {
...
+
+  /**
+    * Compute Shapley feature attributions for a given input in this node's subtree
+    *
+    * @param input for which to compute feature attributions.
+    * @param omitFeatures feature indices to omit in computing Shapley values
+    * @return array of vector-valued attributions for each feature
+    *         One DenseVector[Double] per feature, each of length equal to the output dimension.
+    */
+  def shapley(
```
bfolie, line 26: Why is `shapley` defined at such a high level? It doesn't seem like something that would extend to all models.

src/main/scala/io/citrine/lolo/bags/Bagger.scala
```diff
@@ -280,6 +281,32 @@ class BaggedModel[+T: ClassTag](
...
+    */
+  override def shapley(input: Vector[Any], omitFeatures: Set[Int] = Set()): Option[Vector[DenseVector[Double]]] = {
+    val ensembleShapley = models.map(model => model.shapley(input, omitFeatures))
+    if (!ensembleShapley.head.isDefined) {
+      None
+    }
+    assert(ensembleShapley.forall(x=>x.isDefined))
+    def sumReducer(a: Option[Vector[DenseVector[Double]]],
+                   b: Option[Vector[DenseVector[Double]]]): Option[Vector[DenseVector[Double]]] = {
+      (a ++ b).reduceOption[Vector[DenseVector[Double]]]{case (x,y) => x.zip(y).map{case (v1,v2) => (v1 + v2)}}
```
bfolie, line 302: This seems like it could be written more simply

src/test/scala/io/citrine/lolo/bags/BaggerTest.scala
```diff
@@ -334,6 +352,39 @@ class BaggerTest {
       assert(sigma.forall(_ > 0.0), s"Found an predicted uncertainty of ${sigma.min} during trial $idx")
     }
   }
+
+  /**
+    * Test Shapley values are correctly averaged.
```
bfolie, line 357:
```suggestion
    * Test Shapley values are correctly averaged over trees.
```

src/test/scala/io/citrine/lolo/learners/RandomForestTest.scala
```diff
@@ -161,6 +163,67 @@ class RandomForestTest {
     learner.train(trainingData).getModel()
   }
 
+  def shapleyCompare(
+                      trainingData: Seq[(Vector[Any],Double)],
+                      evalLocation: Vector[Any],
+                      expected: Vector[Double]
+                    ): Unit = {
+    assert(trainingData.length >= 256, "This test is likely to fail randomly when training data are too small.")
```
bfolie, line 171: Is this still the case, given the random seeds?

src/test/scala/io/citrine/lolo/learners/RandomForestTest.scala
```diff
@@ -161,6 +163,67 @@ class RandomForestTest {
...
+      case x: Option[Vector[DenseVector[Double]]] => {
+        val a = x.get
+        assert(a.length == trainingData.head._1.length, "Expected one Shapley value per feature.")
+        assert(a.head.length == 1, "Expected a single output dimension.")
+        a.map{_(0)}
+      }
+      case _ => fail("Unexpected return type.")
+    }
+    expected.zip(actual).foreach {
+      case (e: Double, a: Double) => assert(Math.abs(e - a) < 1)
```
bfolie, line 183: Why a tolerance of 1? It seems like it would depend on the problem.

src/test/scala/io/citrine/lolo/learners/RandomForestTest.scala
```diff
@@ -161,6 +163,67 @@ class RandomForestTest {
...
+      (Vector(1.0, 0.0), 80.0),
+      (Vector(1.0, 0.2), 70.0),
+      (Vector(0.0, 1.0), 0.0),
+      (Vector(0.0, 0.2), 0.0),
+      (Vector(0.0, 0.0), 0.0)
+    )
+    val expected3 = Vector(45.8333333333333, 12.5)
+    shapleyCompare((1 to 64).map(_=>trainingData3).flatten, Vector[Any](1.0, 1.0), expected3)
+
+    // TODO(grobinson): Test case with 5 or more features to exercise all factorial terms in Lundberg Equation 2.
```
bfolie, line 225: Why not copy the `trainingData4` example from `RegressionTreeTest`?

bfolie:
Tests give me confidence that it's working. I had some stylistic and structural questions. I also have not dived in to the details of the algorithm itself (specifically what's going on in `shapleyRecurse`) which I should do, but it's not blocking

src/main/scala/io/citrine/lolo/bags/Bagger.scala
gregor-robinson: Yeah, I'm not thrilled with it. But the option makes things a little annoying, as does  the vector of vectors. If model.shapley returned an `Option[DenseMatrix[Double]]`, then `Seq[DenseMatrix[Double]]` would appear at first glance to be nearly compatible with `breeze.stats.mean`, except that breeze doesn't support tensors :(
Do you have another suggestion?

src/main/scala/io/citrine/lolo/Model.scala
gregor-robinson: It does extend to all models. There's a simple closed form available for linear regression, for example, and a decent kernel approximation for general black-box models.

src/main/scala/io/citrine/lolo/Model.scala
gregor-robinson: This is specified as a breeze vector so that it can be directly used with linalg operations, especially vector addition.

src/test/scala/io/citrine/lolo/bags/BaggerTest.scala
gregor-robinson: Flatten doesn't work in retrieving the doubles because of DenseVector.

src/test/scala/io/citrine/lolo/bags/BaggerTest.scala
gregor-robinson: Good call.

src/test/scala/io/citrine/lolo/learners/RandomForestTest.scala
gregor-robinson: Good catch

src/test/scala/io/citrine/lolo/learners/RandomForestTest.scala
gregor-robinson: Was arbitrary and I hadn't patched it up. Thanks.

src/main/scala/io/citrine/lolo/bags/Bagger.scala
bfolie: Yes, but after trying them out in IntelliJ they don't seem to work. I was assuming that `DenseVector` behaved the same as `Vector`.

src/main/scala/io/citrine/lolo/trees/Nodes.scala
```diff
@@ -63,6 +111,85 @@ class InternalModelNode[T <: PredictionResult[Any]](
...
+    }
+
+    if (omitFeatures.contains(parentFeatureIndex)) {
+      hot match {
+        case _: ModelNode[T] | _: ModelLeaf[T] => hot.shapleyRecurse(
+          input, omitFeatures, parentPath, parentZeroFraction*hot.getTrainingWeight()/trainingWeight, parentOneFraction*hot.getTrainingWeight()/trainingWeight, split.getIndex(), importances)
+        case _ => None
+      }
+      cold match {
+        case _: ModelNode[T] | _: ModelLeaf[T] => hot.shapleyRecurse(
```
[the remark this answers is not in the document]
gregor-robinson: `screaming.jpg`

The whole manner of handling missing data was broken. It's fixed now, and this fix was much easier to implement with a newly-pushed pattern that has `shapleyRecurse` return a DenseMatrix that is added pairwise at each recursive step.

gregor-robinson:
@maxhutch some very minor edits. I think this is ready to merge.

PR #202, opened 2020-02-28
Remove bias for duplicate input rows with different categorical labels.

gregor-robinson:
Lolo classification forests using the default `GuessTheMeanLearner` do something that I find problematic when there are duplicated input rows in the training data. In particular, if `ClassificationTree` is trained on data in which the input `I` maps to both class labels `L1` and `L2` in equal proportion, only the label that appears first in the dataset passed to the leaf learner will affect that leaf's predictions. Although bagging should partially mitigate this effect, it will still bias predictions near `I` toward the categories that appear earlier in the training data whenever there are multiple data that have inputs `I` associated with different labels. This bug-report-as-test-case demonstrates this bias in the case of a classification forest.

### DIFF
--- a/src/main/scala/io/citrine/lolo/linear/GuessTheMean.scala
+++ b/src/main/scala/io/citrine/lolo/linear/GuessTheMean.scala
@@ -20,7 +20,7 @@ case class GuessTheMeanLearner() extends Learner {
     val data = trainingData.map(_._2).zip(weights.getOrElse(Seq.fill(trainingData.size)(1.0)))
     val mean = data.head._1 match {
       case x: Double => data.asInstanceOf[Seq[(Double, Double)]].map(p => p._1 * p._2).sum / data.map(_._2).sum
-      case x: Any => Random.shuffle(data.groupBy(_._1).mapValues(_.map(_._2).sum)).maxBy(_._2)._1
+      case x: Any => Random.shuffle(data.groupBy(_._1).mapValues(_.map(_._2).sum).toSeq).maxBy(_._2)._1
     }
 
     new GuessTheMeanTrainingResult(new GuessTheMeanModel(mean))
--- a/src/main/scala/io/citrine/lolo/linear/GuessTheMean.scala
+++ b/src/main/scala/io/citrine/lolo/linear/GuessTheMean.scala
@@ -2,6 +2,8 @@ package io.citrine.lolo.linear
 
 import io.citrine.lolo.{Learner, Model, PredictionResult, TrainingResult}
 
+import scala.util.Random
+
 /**
   * Created by maxhutch on 11/15/16.
   */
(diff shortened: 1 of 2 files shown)


maxhutch:
@gregor-robinson : note that the job failures logged on travis are scalastyle, which gates the execution of the tests.  Can you fix the style so we can see the signal in the tests?

src/test/scala/io/citrine/lolo/learners/RandomForestTest.scala
```diff
@@ -72,6 +73,52 @@ class RandomForestTest {
...
+        (0,0)
+      }
+    }.asInstanceOf[Seq[(Int,Int)]].reduce{(a: (Int,Int),b: (Int,Int))=>(a._1 + b._1, a._2 + b._2)}
+
+    // Posterior beta distribution with Jeffrey prior.
+    val d = new Beta(winsSuffixed + 0.5, winsPrefixed + 0.5)
+    val l = d.inverseCdf(2e-6)
+    val r = d.inverseCdf(1-2e-6)
+    val tol = 1e-2
+    // println(f"Rate=${d.mean}%.3f (1e-6 CI ${l}%.3f - ${r}%.3f)")
```
bfolie: Remove?

gregor-robinson: I intentionally put it there as a breadcrumb for future inspection, but I can remove if you don't like it.

gregor-robinson: On second thought, I'll remove it since I put the CI in the error messages. Also I want to fix the spelling of Jeffreys prior :)

src/main/scala/io/citrine/lolo/linear/GuessTheMean.scala
```diff
@@ -18,7 +20,7 @@ case class GuessTheMeanLearner() extends Learner {
     val data = trainingData.map(_._2).zip(weights.getOrElse(Seq.fill(trainingData.size)(1.0)))
     val mean = data.head._1 match {
       case x: Double => data.asInstanceOf[Seq[(Double, Double)]].map(p => p._1 * p._2).sum / data.map(_._2).sum
-      case x: Any => data.groupBy(_._1).mapValues(_.map(_._2).sum).maxBy(_._2)._1
+      case x: Any => Random.shuffle(data.groupBy(_._1).mapValues(_.map(_._2).sum)).maxBy(_._2)._1
```
maxhutch:
```suggestion
      case x: Any => Random.shuffle(data.groupBy(_._1).mapValues(_.map(_._2).sum).toSeq).maxBy(_._2)._1
```

I'm a bit uncomfortable relying on the stability of the order of a `MapValues`.  `toSeq` makes the order a bit more explicit.

Issue #206, opened 2020-03-06
Let users pass an RNG object to nondeterministic functions


gregor-robinson:
For the sake of reproducible results, including deterministic unit and integration testing, Bagger and GuessTheMeanLearner should have access to a user-specified RNG object (at least at training time). It would be possible to expose an interface that allows users to specify a RNG seed, but that is strictly less powerful and prevents the use of desirable alternative RNGs such as those that are counter-based. Any other nondeterministic functionality should similarly accept an RNG object, so please note any other nondeterminism to keep in mind -- cross validation, for example.

gregor-robinson:
Heavy lift would be gained by doing this just for bag creation, which can be standardized in BaggerHelper (see https://github.com/CitrineInformatics/lolo/issues/212)

mrupp-citrine:
👍 for control of pseudo-random numbers. But why is it necessary to pass the generator object? A seed should be fine _as long as all the following results are deterministic_. This would be the case for example if an instance of the NumPy generator would be created locally. That is not the most efficient solution (this might be a very efficient approach: https://github.com/google/jax/blob/master/design_notes/prng.md) but easy to implement.

gregor-robinson:
It's good practice to expose an option to pass a generator rather than a seed because it allows the user to swap the RNG with a different algorithm. This helps make code more testable, instrumentable, and flexible --- and it's negligible implementation cost.

mrupp-citrine:
I would argue that passing a generator increases interface complexity but provides limited benefit except for cases where the effect of the generator itself should be tested. The latter is unlikely to become a priority, and can be (partially) done with specialized test suites such as the diehard tests.

In the unlikely case that passing a generator should become necessary, the seed argument could be upgraded to int or generator, using a default generator if only the seed is specified. So passing a seed does not preclude later passing of generators.

bfolie:
`Bagger`, `GuessTheMeanLearner`, `RegressionTreeLearner`, `ClassificationTreeLearner`, `MultiTaskTree`, `RegressionSplitter`, `ClassificationSplitter`, `MultiTaskSplitter`, `RandomForest`, and `ExtraRandomTrees` all have access to random number generators, so I'm closing this.

I noticed one place where the `rng` is not passed through in `RandomForest`, so I fixed that in #258.

Our random forest still isn't fully reproducible because the trees are trained in parallel. I opened an issue: #259

PR #210, opened 2020-03-12
Fix uncertainty interface mixup

gregor-robinson:
* Check `observational` parameter of `getUncertainty`
* Make `getStdDevMean` a standard deviation instead of a variance
* Nudge `MultiTaskBagger` UQ interface toward that of `Bagger`.
* Add `BaggerUtil` so functionality common to `Bagger` and `MultiTaskBagger` can be maintained in one place. (Currently used only by `MultiTaskBagger`, and a subsequent PR will swing over functionality from `Bagger`.)

Note that the second two items, while important, started with a wild goose chase because of nondeterministic tests (see https://github.com/CitrineInformatics/lolo/issues/206)

### DIFF
--- a/src/main/scala/io/citrine/lolo/bags/BaggedResult.scala
+++ b/src/main/scala/io/citrine/lolo/bags/BaggedResult.scala
@@ -57,17 +57,24 @@ case class BaggedSingleResult(
   override def getExpected(): Seq[Double] = Seq(expected)
 
   private lazy val expected = treePredictions.sum / treePredictions.length
-  private lazy val treeVariance: Double = treePredictions.map(x => Math.pow(x - expected, 2.0)).sum / treePredictions.length
+  private lazy val treeVariance: Double = {
+    assert(treePredictions.length > 1, "Bootstrap variance undefined for fewer than 2 bootstrap samples." + treePredictions.length)
+    treePredictions.map(x => Math.pow(x - expected, 2.0)).sum / treePredictions.length
+  }
 
-  override def getStdDevObs(): Option[Seq[Double]] = Some(Seq(treeVariance))
+  override def getStdDevObs(): Option[Seq[Double]] = Some(Seq(scalarUncertainty))
 
-  override def getStdDevMean(): Option[Seq[Double]] = Some(Seq(scalarUncertainty))
+  override def getStdDevMean(): Option[Seq[Double]] = Some(Seq(Math.sqrt(treeVariance)))
 
   /**
    * For the sake of parity, we were using this method
    */
   override def getUncertainty(observational: Boolean): Option[Seq[Any]] = {
-    getStdDevMean()
+    if (observational) {
+      getStdDevObs()
+    } else {
+      getStdDevMean()
+    }
   }
 
   private lazy val scalarUncertainty: Double = {
@@ -177,19 +184,24 @@ case class BaggedMultiResult(
     */
   override def getExpected(): Seq[Double] = expected
 
-  override def getStdDevObs(): Option[Seq[Double]] = Some{
-    expectedMatrix.asInstanceOf[Seq[Seq[Double]]].zip(expected.asInstanceOf[Seq[Double]]).map{case (b, y) =>
-      b.map{x => Math.pow(x - y, 2.0)}.sum / b.size
+  override def getStdDevObs(): Option[Seq[Double]] = Some(uncertainty)
+
+  override def getStdDevMean(): Option[Seq[Double]] = Some{
+    expectedMatrix.asInstanceOf[Seq[Seq[Double]]].zip(expected.asInstanceOf[Seq[Double]]).map { case (b, y) =>
+      assert(b.size > 2)
+      Math.sqrt(b.map { x => Math.pow(x - y, 2.0) }.sum / (b.size - 1))
     }
   }
-
-  override def getStdDevMean(): Option[Seq[Double]] = Some(uncertainty)
 
   /**
    * For the sake of parity, we were using this method
    */
   override def getUncertainty(observational: Boolean): Option[Seq[Any]] = {
-    getStdDevMean()
+    if (observational) {
+      getStdDevObs()
+    } else {
+      getStdDevMean()
+    }
   }
 
   /**
--- a/src/main/scala/io/citrine/lolo/bags/BaggedResult.scala
+++ b/src/main/scala/io/citrine/lolo/bags/BaggedResult.scala
@@ -64,7 +64,7 @@ case class BaggedSingleResult(
 
   override def getStdDevObs(): Option[Seq[Double]] = Some(Seq(scalarUncertainty))
 
-  override def getStdDevMean(): Option[Seq[Double]] = Some(Seq(Math.sqrt(treeVariance)))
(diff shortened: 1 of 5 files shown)


src/main/scala/io/citrine/lolo/bags/BaggedResult.scala
```diff
@@ -57,17 +57,24 @@ case class BaggedSingleResult(
...
+  private lazy val treeVariance: Double = {
+    assert(treePredictions.length > 1, "Bootstrap variance undefined for fewer than 2 bootstrap samples." + treePredictions.length)
+    treePredictions.map(x => Math.pow(x - expected, 2.0)).sum / treePredictions.length
+  }
 
-  override def getStdDevObs(): Option[Seq[Double]] = Some(Seq(treeVariance))
+  override def getStdDevObs(): Option[Seq[Double]] = Some(Seq(scalarUncertainty))
 
-  override def getStdDevMean(): Option[Seq[Double]] = Some(Seq(scalarUncertainty))
+  override def getStdDevMean(): Option[Seq[Double]] = Some(Seq(Math.sqrt(treeVariance)))
```
maxhutch: My benchmarking suggested that `Math.sqrt(treeVariance) * rescale` would be `getStdDevObs` not `getStdDevMean`.  Do we know that it goes to zero as the number of samples of a noisy signal increases?

src/main/scala/io/citrine/lolo/bags/BaggedResult.scala
```diff
@@ -177,19 +184,24 @@ case class BaggedMultiResult(
...
 
-  override def getStdDevObs(): Option[Seq[Double]] = Some{
-    expectedMatrix.asInstanceOf[Seq[Seq[Double]]].zip(expected.asInstanceOf[Seq[Double]]).map{case (b, y) =>
-      b.map{x => Math.pow(x - y, 2.0)}.sum / b.size
+  override def getStdDevObs(): Option[Seq[Double]] = Some(uncertainty)
+
+  override def getStdDevMean(): Option[Seq[Double]] = Some{
+    expectedMatrix.asInstanceOf[Seq[Seq[Double]]].zip(expected.asInstanceOf[Seq[Double]]).map { case (b, y) =>
+      assert(b.size > 2)
+      Math.sqrt(b.map { x => Math.pow(x - y, 2.0) }.sum / (b.size - 1))
```
maxhutch: Likewise here.

src/main/scala/io/citrine/lolo/bags/BaggedResult.scala
```diff
@@ -57,17 +57,24 @@ case class BaggedSingleResult(
   override def getExpected(): Seq[Double] = Seq(expected)
 
   private lazy val expected = treePredictions.sum / treePredictions.length
-  private lazy val treeVariance: Double = treePredictions.map(x => Math.pow(x - expected, 2.0)).sum / treePredictions.length
+  private lazy val treeVariance: Double = {
+    assert(treePredictions.length > 1, "Bootstrap variance undefined for fewer than 2 bootstrap samples." + treePredictions.length)
+    treePredictions.map(x => Math.pow(x - expected, 2.0)).sum / treePredictions.length
```
maxhutch: No bessel correction?

src/main/scala/io/citrine/lolo/bags/BaggerUtil.scala
```diff
@@ -0,0 +1,58 @@
+package io.citrine.lolo.bags
+
+import io.citrine.lolo.util.Async
+import io.citrine.lolo.{Model, PredictionResult}
+
+import scala.collection.parallel.immutable.ParSeq
+
+case class BaggerHelper(
```
maxhutch: Docstring?  Should this class be protected?

src/test/scala/io/citrine/lolo/bags/BaggedResultTest.scala
```diff
@@ -31,6 +31,50 @@ class BaggedResultTest {
...
+        }
+
+        sigmaObs.zip(sigmaMean).foreach { case (sObs, sMean) => assert(sObs > sMean, "Uncertainty should be greater when observational = true.") }
+
+        // We have strong theoretical guarantees on the behavior of GuessTheMeanLearner, so let's exercise them.
+        // TODO(grobinson): implement similar test for RegressionTreeLearner.
+        if (baseLearner.isInstanceOf[GuessTheMeanLearner]) {
+          // NOTE: these bounds reflect a ~3x systematic variance under-estimation in this particular test setting.
+          val rtolLower = 5.0  // Future recalibration should decrease this number.
+          val rtolUpper = 1.0  // Future recalibration should increase this number.
```
bfolie: Decrease? Or I think the assert statement should be changed to `s * rTolUpper < noiseLevel`. Currently increasing `rTolUpper` makes the assertion weaker

src/main/scala/io/citrine/lolo/bags/BaggerUtil.scala
maxhutch: This is only used in the MultiTaskBagger; did you mean to use it in the Bagger too?

src/test/scala/io/citrine/lolo/bags/BaggedResultTest.scala
gregor-robinson: Future recalibration should make both just slightly more than 1. It's set to 1 now to catch changes, because it's currently underestimating consistently, but a well-calibrated estimate would have both around, say, 2.

src/main/scala/io/citrine/lolo/bags/BaggedResult.scala
gregor-robinson: I have observed that it goes to zero in the number of *duplicated* samples of a noisy signal (i.e. observations with the same inputs) but things get harder to analyze with more complicated base learners. We'd still be able to analyze desired behavior of a linear base learner, but that's not much different from GuessTheMean.

src/main/scala/io/citrine/lolo/bags/BaggedResult.scala
gregor-robinson: I removed the Bessel correction because it was inconsistent with BaggedMultiResult and couples to the Jackknife in a way that will need to be analyzed more closely. I have opened an issue for this: https://github.com/CitrineInformatics/lolo/issues/213

src/main/scala/io/citrine/lolo/bags/BaggerUtil.scala
gregor-robinson: Documented and protected.
I intend to use this in Bagger too, but I want to do that cleanup in another PR to get this out the door.

src/test/scala/io/citrine/lolo/bags/BaggedResultTest.scala
bfolie: OK, so `sigmaObs` is currently under-estimating, and `s < 1 * noiseLevel` is reliably true now, but if it were working correctly a more appropriate bound would be `s < 2 * noiseLevel`? And you're leaving it at 1 so you can observe the test "failing" when sigmaObs is re-calibrated, at which point you'll increase `rTolUpper`? I think that makes sense.

Issue #213, opened 2020-03-12
Unbias variance estimator


gregor-robinson:
getStdDevMean currently uses a biased variance estimator. This should be unbiased by replacing the denominator with `treePredictions.length - 1`

gregor-robinson:
@mrupp-citrine made the good point in reviewing #217 that we should care more about debiasing the standard deviation estimate. So, although #216 adds a Bessel correction, we should at least use an N-3/2 correction so that the standard deviation is unbiased to second order.

gregor-robinson:
@mrupp-citrine suggests the following, which is within scope of this issue:
> The above matters only for small ensembles. Still, documenting (in source comments) the answers to these questions might help later to better remember why these decisions (e.g., to use Bessel correction) were made.

Issue #207, opened 2020-03-09
Change in predictive uncertainties from lolopy v1.0.4 to v1.1.0


mrupp-citrine:
Train a `lolo` random forest on `n=100` training points `(x,y)`, where `x` from -5 to 5 and `y(x) = f(x) + eps`, with `f(x) = 2x+1` and `eps ~ N(0,1)`.

In `lolopy` version `1.0.4`, the mean predictive uncertainty is around 1, matching the aleatoric uncertainty / noise.

In the next published `lolopy` version 1.1.0, mean predictive uncertainties are around 0.5, which seems too low given the noise.

What changed, and why?

Here is an example with `smlb`:

```
n, m, xlen = 100, 600, 10
train_inputs = np.reshape(np.linspace(-xlen / 2, +xlen / 2, n), (n, 1))
train_labels = (train_inputs * 2 + 1).flatten()
train_data = smlb.TabularLabeledData(data=train_inputs, labels=train_labels)
train_data = smlb.LabelNoise(noise=smlb.NormalNoise(rng=0)).fit(train_data).apply(train_data)

valid_inputs = np.reshape(np.linspace(-xlen / 2, +xlen / 2, m), (m, 1))
valid_labels = (valid_inputs * 2 + 1).flatten()
valid_data = smlb.TabularLabeledData(data=valid_inputs, labels=valid_labels)
valid_data = smlb.LabelNoise(noise=smlb.NormalNoise(rng=1)).fit(valid_data).apply(valid_data)

rf = RandomForestRegressionLolo()
preds = rf.fit(train_data).apply(valid_data)
mae = smlb.MeanAbsoluteError().evaluate(valid_data.labels(), preds)

# for perfect predictions, expect MAE of 1.12943
# (absolute difference between draws from two unit normal distributions)
assert np.allclose(mae, 1.13, atol=0.25)
assert np.allclose(np.median(preds.stddev), 1, atol=0.5)
```

gregor-robinson:
Fix: https://github.com/CitrineInformatics/lolo/pull/210

gregor-robinson:
Fixed deployed https://github.com/CitrineInformatics/lolo/releases/tag/v3.0.1

PR #216, opened 2020-03-18
UQ sanity check

gregor-robinson:
This PR adds a sanity check on reported standard deviations for regression, and re-applies the conversion to using the rescaled bootstrap ensemble for predictive uncertainty and jackknife for standard error. This PR is a step toward fixing https://github.com/CitrineInformatics/lolo/issues/207 and also implements the Bessel correction (https://github.com/CitrineInformatics/lolo/issues/213).

### DIFF
--- a/src/test/scala/io/citrine/lolo/bags/BaggedResultTest.scala
+++ b/src/test/scala/io/citrine/lolo/bags/BaggedResultTest.scala
@@ -31,44 +31,71 @@ class BaggedResultTest {
     }
   }
 
+  /**
+    * Test that uncertainty estimates are within reasonable bounds.
+    */
   @Test
   def testBaggedSingleResultGetUncertainty(): Unit = {
     val noiseLevel = 100.0
     val rng = new Random(237485L)
-    Seq(RegressionTreeLearner(),GuessTheMeanLearner()).foreach{ baseLearner =>
-      Seq(30,100,301).foreach { nRows =>
-        val trainingDataTmp = TestUtils.generateTrainingData(nRows, 1, noise = 0.0, function = _ => 0.0, seed = rng.nextLong())
-        val trainingData = (trainingDataTmp).map { x => (x._1, x._2 + noiseLevel * rng.nextDouble()) }
-        val baggedLearner = Bagger(baseLearner, numBags = 2 * nRows, uncertaintyCalibration = true)
-        val RFMeta = baggedLearner.train(trainingData)
-        val RF = RFMeta.getModel()
-        val results = RF.transform(trainingData.take(1).map(_._1))
 
-        val sigmaObs: Seq[Double] = results.getUncertainty().get.asInstanceOf[Seq[Double]]
-//        sigmaMean.zip(results.asInstanceOf[RegressionResult].getStdDevMean().get).foreach{ case (a,b) =>
-//          assert(a == b, "Expected getUncertainty(observational=false)=getStdDevMean()")
-//        }
-//        val sigmaMean: Seq[Double] = results.getUncertainty(observational = false).get.asInstanceOf[Seq[Double]]
-        sigmaObs.zip(results.asInstanceOf[RegressionResult].getStdDevObs().get).foreach{ case (a,b) =>
-          assert(a == b, "Expected getUncertainty()=getStdDevObs()")
-        }
+    Seq(RegressionTreeLearner(), GuessTheMeanLearner()).foreach { baseLearner =>
+      // These are in Seqs as a convenience for repurposing this test as a diagnostic tool.
+      Seq(128).foreach { nRows =>
+        Seq(16).foreach { nCols =>
+          Seq(2).map { n => n * nRows }.foreach { nBags =>
+            // Used for error output.
+            val configDescription =s"learner=${baseLearner.getClass().toString()}\tnRows=$nRows\tnCols=$nCols\tnumBags=$nBags"
 
-//        sigmaObs.zip(sigmaMean).foreach { case (sObs, sMean) => assert(sObs > sMean, "Uncertainty should be greater when observational = true.") }
+            // Count times getStdDevObs is > and < getStdDevMean
+            var countStdObsGtStdMean = 0.0
+            var countStdObsLtStdMean = 0.0
 
-        // We have strong theoretical guarantees on the behavior of GuessTheMeanLearner, so let's exercise them.
-        // TODO(grobinson): implement similar test for RegressionTreeLearner.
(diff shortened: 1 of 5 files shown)


src/test/scala/io/citrine/lolo/bags/BaggedResultTest.scala
```diff
@@ -31,44 +31,71 @@ class BaggedResultTest {
...
-          sigmaObs.foreach { s =>
-            assert(rtolLower * s > noiseLevel, "Observational StdDev getUncertainty() is too small.")
-            assert(s < rtolUpper * noiseLevel, "Observational StdDev getUncertainty() is too large.")
+    Seq(RegressionTreeLearner(), GuessTheMeanLearner()).foreach { baseLearner =>
+      // These are in Seqs as a convenience for repurposing this test as a diagnostic tool.
+      Seq(128).foreach { nRows =>
+        Seq(16).foreach { nCols =>
+          Seq(2).map { n => n * nRows }.foreach { nBags =>
+            // Used for error output.
+            val configDescription =s"learner=${baseLearner.getClass().toString()}\tnRows=$nRows\tnCols=$nCols\tnumBags=$nBags"
```
maxhutch:
```suggestion
            val configDescription = s"learner=${baseLearner.getClass().toString()}\tnRows=$nRows\tnCols=$nCols\tnumBags=$nBags"
```

src/test/scala/io/citrine/lolo/bags/BaggedResultTest.scala
```diff
@@ -31,44 +31,71 @@ class BaggedResultTest {
...
+                if (sObs > sMean) {
+                  countStdObsGtStdMean += 1
+                } else {
+                  countStdObsLtStdMean += 1
+                }
+              }
+
+              // We have strong theoretical guarantees on the behavior of GuessTheMeanLearner, so let's exercise them.
+              // NOTE: these bounds reflect a ~3x systematic variance under-estimation in this particular test setting.
+              var rtolLower = if (baseLearner.isInstanceOf[GuessTheMeanLearner]) 3.5 else 10.0 // Future recalibration should decrease this number.
```
maxhutch: Instead of defining these as `var` and re-assigning, you could wrap the whole `sigmaObs` check in brackets to define a sub-scope and use `val`.  Or, you could move this logic into a function.

maxhutch: Also, `match` is generally preferred over `isInstanceOf`:
```
val rtolLower = baseLearner match {
  case _: GuessTheMeanLearner => 3.5
  case _: Any => 10.0
}
```

gregor-robinson:
> I strongly suggest making this call explicit in that it is `getStdDevObs`:
> https://github.com/CitrineInformatics/lolo/blob/bugfix/uq-sanity-check/src/main/scala/io/citrine/lolo/bags/Bagger.scala#L115
> and here:
> https://github.com/CitrineInformatics/lolo/blob/bugfix/uq-sanity-check/src/main/scala/io/citrine/lolo/bags/BaggerUtil.scala#L45

The problem here is that getUncertainty applies both to regression and also to classification, while getStdDevObs is only a regression concept. I propose that we merge this to unblock hotfixes without trying to unify the interfaces for now.

maxhutch:
> > I strongly suggest making this call explicit in that it is `getStdDevObs`:
> > https://github.com/CitrineInformatics/lolo/blob/bugfix/uq-sanity-check/src/main/scala/io/citrine/lolo/bags/Bagger.scala#L115
> > and here:
> > https://github.com/CitrineInformatics/lolo/blob/bugfix/uq-sanity-check/src/main/scala/io/citrine/lolo/bags/BaggerUtil.scala#L45
> 
> The problem here is that getUncertainty applies both to regression and also to classification, while getStdDevObs is only a regression concept. I propose that we merge this to unblock hotfixes without trying to unify the interfaces for now.

The logic in that block only applies to regression problems anyways.  We could do:
```
  val uncertainty = predicted match {
    case x: RegressionResult => x.getStdDevObs().get.head
    case x: Any => throw ....
}
```

Or slap a TODO on it.

gregor-robinson:
> The logic in that block only applies to regression problems anyways.

Oh, good call. Updated.

src/test/scala/io/citrine/lolo/bags/BaggedResultTest.scala
```diff
@@ -31,44 +31,87 @@ class BaggedResultTest {
...
+        Seq(16).foreach { nCols =>
+          Seq(2).map { n => n * nRows }.foreach { nBags =>
+            // Used for error output.
+            val configDescription = s"learner=${baseLearner.getClass().toString()}\tnRows=$nRows\tnCols=$nCols\tnumBags=$nBags"
+
+            // Count times getStdDevObs is > and < getStdDevMean
+            var countStdObsGtStdMean = 0.0
+            var countStdObsLtStdMean = 0.0
+
+            (1 to 10).foreach { _ =>
```
bfolie: Not blocking, but in order to avoid `var` you could do
```
val sigmaObsAndMean: Seq[Seq[(Double, Double)]] = (1 to 10).map {_ =>
```
then use `flatMap` to turn it into a `Seq[(Double, Double)]` and use `.count` to count the number for which `sObs > sMean`. You could also do the bounds checks on this sequence.

src/test/scala/io/citrine/lolo/bags/MultiTaskBaggerTest.scala
```diff
@@ -45,45 +45,88 @@ class MultiTaskBaggerTest {
...
+                } else {
+                  countStdObsLtStdMean += 1
+                }
+              }
+
+              // We have strong theoretical guarantees on the behavior of GuessTheMeanLearner, so let's exercise them.
+              // NOTE: these bounds reflect a ~3x systematic variance under-estimation in this particular test setting.
+              {
+                val rtolLower = baseLearner match {
+                  case _: GuessTheMeanLearner => 3.5
```
bfolie: Did you intend to include `GuessTheMeanLearner` in the sequence of base learners?

gregor-robinson: I would like to, but there's no such MultiTaskLearner. I cleaned this up to be a little more explicit about why the code looks like this when there's just one baseLearner.

src/test/scala/io/citrine/lolo/bags/BaggedResultTest.scala
gregor-robinson: Cool, done. (Mostly to improve my Scala idioms.)

PR #217, opened 2020-03-24
Release/3.0.1

gregor-robinson:
# Version 3.0.1 is released!

### Fixes
* Release 3.0.0 lacked a rescaling factor and a square root in getStdDevObs. Version 3.0.1 fixes this, improves test coverage, and introduces a Bessel correction.


mrupp-citrine:
Two questions:

The Bessel correction (dividing by `n-1` instead of by `n`) removes bias at the expense of increased mean squared error. So, the estimate will have lower bias but higher variance. 

We use the Bessel correction to estimate the tree ensemble variance. Why is lower-bias higher-variance desirable here?

The Bessel correction results in an unbiased estimator of variance, but not of the standard deviation.

In line 97, the Bessel correction is removed from the estimate.
In line 62, when computing `treeVariance`, it is included.
In line 83, when computing `stdDevObs`, it is not removed (`sqrt(treeVariance)`).

Is this intentional, and if yes, why does line 83 require no correction?

The above matters only for small ensembles. Still, documenting (in source comments) the answers to these questions might help later to better remember why these decisions (e.g., to use Bessel correction) were made.

mrupp-citrine:
Kristin Lennox made this comment in the tech-dev channel:

> Regression trees are estimators of the conditional mean of the data, so taking the variance across trees will underestimate the variance of the data.

Is this relevant to the estimates returned here?

gregor-robinson:
> Two questions:
> 
> The Bessel correction (dividing by `n-1` instead of by `n`) removes bias at the expense of increased mean squared error. So, the estimate will have lower bias but higher variance.
> 
> We use the Bessel correction to estimate the tree ensemble variance. Why is lower-bias higher-variance desirable here?

The bias is a small but systematic underestimation, and in practice it is generally more dangerous to underestimate uncertainty than to overestimate it.

> The Bessel correction results in an unbiased estimator of variance, but not of the standard deviation.

That's a great point -- I was correcting the variance without thinking about the ultimate use case as a standard deviation. Since we're presenting a standard deviation, it's better to use a denominator of N-3/2 so that the bias decays quadratically instead of linearly. However, the Bessel correction does still partially remove the bias in standard deviation. In the interest of moving this release forward, since this is already an improvement, I propose merging this. I've updated #213 and will keep it open so that doesn't get lost in the cracks, but I can open a PR against develop as soon as I finish this sprint's other work.

> In line 97, the Bessel correction is removed from the estimate.
> In line 62, when computing `treeVariance`, it is included.
> In line 83, when computing `stdDevObs`, it is not removed (`sqrt(treeVariance)`).
> Is this intentional, and if yes, why does line 83 require no correction?

This is intentional because AFAICT the jackknife approach is formulated with an N^2 in the denominator. (I think I found evidence of this is Wager, Hastie, Efron 2014 but the equation isn't jumping out at me right now.) I'm still tracing through the winding threads of literature, with the intent of documenting the specific mathematical/empirical pedigree of this code, so the best I can do on the jackknife component of this is to keep it as it was. Hence removing the variance correction. In hindsight, it would be cleaner to compute the population standard deviation in treeVariance and then correct it at the moment it's required, but I'll save that for work against #217. 

> The above matters only for small ensembles. Still, documenting (in source comments) the answers to these questions might help later to better remember why these decisions (e.g., to use Bessel correction) were made.

Yes, excellent point. I'll add this to #213 as well.

gregor-robinson:
> Kristin Lennox made this comment in the tech-dev channel:
> 
> > Regression trees are estimators of the conditional mean of the data, so taking the variance across trees will underestimate the variance of the data.
> 
> Is this relevant to the estimates returned here?

Yes! This is the purpose of the empirical recalibration: to inflate the bootstrap variance so that it can approximate predictive variance more accurately. This was the conclusion of @maxhutch's recalibration work. There's still ample room for research there, but this is our state of the art for our use cases.

RELEASE.md
```diff
@@ -0,0 +1,20 @@
+# v3.0 is released!
```
maxhutch: @gregor-robinson did you mean to add the 3.0.1 release in this file too?

RELEASE.md
```diff
@@ -0,0 +1,20 @@
...
+
+Lolo version 3.0 is released! Notably, this release introduces a shiny new interface for uncertainty quantification that provides new and improved granular uncertainty quantification (UQ). This release also introduces two new random forest features, and improves the accuracy of predictions and UQ on certain types of machine learning problems.
+
+### What’s New
+
+* Now you can distinguish estimated uncertainty in observations from uncertainty in the mean! You can query each of these with a new interface that exposes granular estimates of the [bias-variance-noise decomposition](https://en.wikipedia.org/wiki/Bias%E2%80%93variance_tradeoff). (#198)
+* A new `Splitter` interface allowing custom tree splitting protocols such as the new `BoltzmannSplitter` used to train [Boltzmann trees](https://www.youtube.com/watch?v=wWChMOkNlWk). (#95)
```
maxhutch: I didn't know this was on youtube :-(

RELEASE.md
gregor-robinson: I added a note about the fix in 3.0.1 as part of 3.0 https://github.com/CitrineInformatics/lolo/pull/217/files/8510e4e64d80d761812dabff12813ce02cb96aaa#diff-781a53e648f3df8d16a08ec083b04bf4R20 -- just trying to reduce clutter.

PR #220, opened 2020-03-31
Add extra random trees

gregor-robinson:
This PR adds extremely randomized trees.

Perhaps the most controversial aspect of this PR is the addition of a new test dependency, Mockito, which is used here as a means of testing a stochastic function. Although it would be possible to simplify an equivalent test of the public interface by resetting the RNG seed twice, mocking the RNG reduces the risk of conveying misleading information to a developer. It is possible to correctly modify the code under test which invalidates the test case, such as reversing the order of calls to the RNG; mocking helps flag the test case itself as the item that must be addressed.

### DIFF
--- a/python/lolopy/learners.py
+++ b/python/lolopy/learners.py
@@ -257,7 +257,7 @@ class RandomForestMixin(BaseLoloLearner):
         """Initialize the RandomForest
 
         Args:
-            num_trees (int): Number of trees to use in the forest
+            num_trees (int): Number of trees to use in the forest (default of -1 sets the number of trees to the number of training rows)
             use_jackknife (bool): Whether to use jackknife based variance estimates
             bias_learner (BaseLoloLearner): Algorithm used to model bias (default: no model)
             leaf_learner (BaseLoloLearner): Learner used at each leaf of the random forest (default: GuessTheMean)
@@ -330,7 +330,7 @@ class ExtraRandomTreesMixIn(BaseLoloLearner):
         """Initialize the ExtraRandomTrees ensemble
 
         Args:
-            num_trees (int): Number of trees to use in the forest
+            num_trees (int): Number of trees to use in the forest (default of -1 sets the number of trees to the number of training rows)
             use_jackknife (bool): Whether to use jackknife based variance estimates
             bias_learner (BaseLoloLearner): Algorithm used to model bias (default: no model)
             leaf_learner (BaseLoloLearner): Learner used at each leaf of the random forest (default: GuessTheMean)
--- a/python/lolopy/learners.py
+++ b/python/lolopy/learners.py
@@ -3,6 +3,7 @@ from abc import abstractmethod, ABCMeta
 import numpy as np
 from lolopy.loloserver import get_java_gateway
 from lolopy.utils import send_feature_array, send_1D_array
+import random
 from sklearn.base import BaseEstimator, RegressorMixin, ClassifierMixin, is_regressor
 from sklearn.exceptions import NotFittedError
 
@@ -253,7 +254,7 @@ class RandomForestMixin(BaseLoloLearner):
     def __init__(self, num_trees=-1, use_jackknife=True, bias_learner=None,
                  leaf_learner=None, subset_strategy="auto", min_leaf_instances=1,
                  max_depth=2**30, uncertainty_calibration=False, randomize_pivot_location=False,
-                 randomly_rotate_features=False):
+                 randomly_rotate_features=False, random_seed=None):
         """Initialize the RandomForest
 
         Args:
@@ -273,6 +274,7 @@ class RandomForestMixin(BaseLoloLearner):
             uncertainty_calibration (bool): whether to re-calibrate the predicted uncertainty based on out-of-bag residuals
             randomize_pivot_location (bool): whether to draw pivots randomly or always select the midpoint
             randomly_rotate_features (bool): whether to randomly rotate real features for each tree in the forest
+            random_seed (int): random number generator seed used for nondeterministic functionality
         """
         super(RandomForestMixin, self).__init__()
 
@@ -287,8 +289,11 @@ class RandomForestMixin(BaseLoloLearner):
         self.uncertainty_calibration = uncertainty_calibration
         self.randomize_pivot_location = randomize_pivot_location
         self.randomly_rotate_features = randomly_rotate_features
+        self.random_seed = random_seed
 
     def _make_learner(self):
+        rng = self.gateway.jvm.scala.util.Random() if self.random_seed is None else self.gateway.jvm.scala.util.Random(self.random_seed)
+
         #  TODO: Figure our a more succinct way of dealing with optional arguments/Option values
         #  TODO: that ^^, please
         learner = self.gateway.jvm.io.citrine.lolo.learners.RandomForest(
@@ -304,7 +309,8 @@ class RandomForestMixin(BaseLoloLearner):
             self.max_depth,
             self.uncertainty_calibration,
             self.randomize_pivot_location,
-            self.randomly_rotate_features
+            self.randomly_rotate_features,
+            rng
         )
         return learner
 
@@ -389,7 +395,7 @@ class ExtraRandomTreesRegressor(BaseLoloRegressor, RandomForestMixin):
 class RegressionTreeLearner(BaseLoloRegressor):
     """Regression tree learner, based on the RandomTree algorithm."""
 
-    def __init__(self, num_features=-1, max_depth=30, min_leaf_instances=1, leaf_learner=None):
+    def __init__(self, num_features=-1, max_depth=30, min_leaf_instances=1, leaf_learner=None, random_seed=None):
         """Initialize the learner
 
         Args:
@@ -403,6 +409,7 @@ class RegressionTreeLearner(BaseLoloRegressor):
         self.max_depth = max_depth
         self.min_leaf_instances = min_leaf_instances
         self.leaf_learner = leaf_learner
+        self.random_seed = random_seed
 
     def _make_learner(self):
         if self.leaf_learner is None:
@@ -420,9 +427,11 @@ class RegressionTreeLearner(BaseLoloRegressor):
             "$lessinit$greater$default$5"
         )()
 
+        rng = self.gateway.jvm.scala.util.Random() if self.random_seed is None else self.gateway.jvm.scala.util.Random(self.random_seed)
+
         return self.gateway.jvm.io.citrine.lolo.trees.regression.RegressionTreeLearner(
             self.num_features, self.max_depth, self.min_leaf_instances,
-            leaf_learner, splitter
+            leaf_learner, splitter, rng
         )
 
 class LinearRegression(BaseLoloRegressor):
--- a/python/lolopy/learners.py
+++ b/python/lolopy/learners.py
@@ -6,7 +6,7 @@ from lolopy.utils import send_feature_array, send_1D_array
 from sklearn.base import BaseEstimator, RegressorMixin, ClassifierMixin, is_regressor
 from sklearn.exceptions import NotFittedError
 
-__all__ = ['RandomForestRegressor', 'RandomForestClassifier']
+__all__ = ['RandomForestRegressor', 'RandomForestClassifier', 'ExtraRandomTreesRegressor']
 
 
 class BaseLoloLearner(BaseEstimator, metaclass=ABCMeta):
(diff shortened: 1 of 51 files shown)


python/lolopy/tests/test_learners.py
```diff
@@ -181,6 +181,51 @@ def test_adjust_rtree_learners(self):
...
+        score = r2_score(y_pred, y)
+        print('R^2:', score)
+        self.assertGreater(score, 0.98)
+
+        # Test with weights (make sure it doesn't crash)
+        rf.fit(X, y, [1.0]*len(y))
+
+        # Make sure feature importances are stored
+        self.assertEqual(np.shape(rf.feature_importances_), (X.shape[1],))
+        self.assertAlmostEqual(1.0, np.sum(rf.feature_importances_))
```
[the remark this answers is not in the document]
gregor-robinson: Despite having a similar name, `get_importance_scores` is intended to capture the impact of training data, which is related but not identical to the gain scores we use to communicate feature importance. I would like to tidy up this terminology, and document it more transparently, but that's not of high relative priority and should be addressed in a separate PR.

src/main/scala/io/citrine/lolo/learners/ExtraRandomTrees.scala
```diff
@@ -0,0 +1,77 @@
...
+  * @param uncertaintyCalibration whether to empirically recalibrate the predicted uncertainties (default: false)
+  * @param randomizePivotLocation whether generate splits randomly between the data points (default: false)
+  * @param randomlyRotateFeatures whether to randomly rotate real features for each tree in the forest (default: false)
+  */
+case class ExtraRandomTrees(
+                             numTrees: Int = -1,
+                             useJackknife: Boolean = true,
+                             biasLearner: Option[Learner] = None,
+                             leafLearner: Option[Learner] = None,
+                             subsetStrategy: Any = "auto",
```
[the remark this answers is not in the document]
gregor-robinson: There are ways to get that functionality, but they're not pretty.
http://milessabin.com/blog/2011/06/09/scala-union-types-curry-howard/

src/main/scala/io/citrine/lolo/trees/splits/ExtraRandomSplitter.scala
```diff
@@ -0,0 +1,137 @@
...
+
+import scala.util.Random
+
+/**
+  * Find the best split for regression problems.
+  *
+  * @param rng random number generator
+  */
+case class ExtraRandomSplitter(
+                               rng: Random = Random
```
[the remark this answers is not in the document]
gregor-robinson: I'm all over it 😎

python/lolopy/tests/test_learners.py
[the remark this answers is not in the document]
mrupp-citrine: Ah, thanks for clarifying. I think the existence of this conversation indicates that more expressive naming is indeed desirable. If not in this PR, could you file a corresponding issue so that this does not get lost?

src/main/scala/io/citrine/lolo/learners/ExtraRandomTrees.scala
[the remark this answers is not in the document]
mrupp-citrine: :lol:
> Obviously it would be nicer if Scala supported union types as primitives
I agree 😄

python/lolopy/tests/test_learners.py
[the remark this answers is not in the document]
gregor-robinson: https://github.com/CitrineInformatics/lolo/issues/224

python/lolopy/learners.py
```diff
@@ -323,6 +323,75 @@ class RandomForestClassifier(BaseLoloClassifier, RandomForestMixin):
...
+                Available options:
+                    "auto": Use the default for lolo (all features for regression; classification not supported)
+                    "log2": Use the base 2 log of the number of features
+                    "sqrt": Use the square root of the number of features
+                    integer: Set the number of features explicitly
+                    float: Use a certain fraction of the features
+            min_leaf_instances (int): Minimum number of features used at each leaf
+            max_depth (int): Maximum depth to which to allow the decision trees to grow
+            uncertainty_calibration (bool): whether to re-calibrate the predicted uncertainty based on out-of-bag residuals
+            randomize_pivot_location (bool): whether to draw pivots randomly or always select the midpoint
```
maxhutch, line 353: This option isn't used for ExtraRandomTrees.

src/main/scala/io/citrine/lolo/learners/ExtraRandomTrees.scala
```diff
@@ -0,0 +1,80 @@
...
+                             minLeafInstances: Int = 1,
+                             maxDepth: Int = Integer.MAX_VALUE,
+                             uncertaintyCalibration: Boolean = false,
+                             randomizePivotLocation: Boolean = false,
```
maxhutch, line 36: This isn't used in the ExtraRandomSplitter.

src/main/scala/io/citrine/lolo/learners/ExtraRandomTrees.scala
```diff
@@ -0,0 +1,80 @@
...
+              case x: String =>
+                println(s"Unrecognized subsetStrategy ${x}; using auto")
+                trainingData.head._1.size
+            }
+          case x: Int =>
+            x
+          case x: Double =>
+            (trainingData.head._1.size * x).toInt
+        }
+        val DTLearner = RegressionTreeLearner(
```
maxhutch, line 64: What about classification?

python/lolopy/tests/test_learners.py
```diff
@@ -181,6 +181,51 @@ def test_adjust_rtree_learners(self):
...
+        y_pred = rf.predict(X)
+        self.assertEqual(len(y_pred), len(y))
+
+        # Test the ability to get importance scores
+        y_import = rf.get_importance_scores(X[:100, :])
+        self.assertEqual((100, len(X)), y_import.shape)
+
+        # Basic test for functionality. R^2 above 0.98 was measured on 27Dec18
+        score = r2_score(y_pred, y)
+        print('R^2:', score)
```
maxhutch, line 209: Can this print be cleaned?

python/lolopy/tests/test_learners.py
```diff
@@ -181,6 +181,51 @@ def test_adjust_rtree_learners(self):
...
+        self.assertGreater(score, 0.98)
```
maxhutch, line 210: This looks like a non-deterministic test.  Plus, I think the threshold was just copied from the RandomForest test; its not clear to me that we'd expect the same from the ExtraRandomTreesRegressor.

gregor-robinson: Good call -- I hadn't updated this since merging in the rng changed. There is still some residual nondeterminism due to threading, so it's not a perfectly deterministic test, but it should be much more reliable now even so.
I have retested the measurements, and they didn't need to change.

src/main/scala/io/citrine/lolo/learners/ExtraRandomTrees.scala
```diff
@@ -0,0 +1,123 @@
...
+  */
+case class ExtraRandomTrees(
+                             numTrees: Int = -1,
+                             useJackknife: Boolean = true,
+                             biasLearner: Option[Learner] = None,
+                             leafLearner: Option[Learner] = None,
+                             subsetStrategy: Any = "auto",
+                             minLeafInstances: Int = 1,
+                             maxDepth: Int = Integer.MAX_VALUE,
+                             uncertaintyCalibration: Boolean = false,
```
maxhutch, line 38: I don't think we can do uncertainty recalibration with bootstrapping disabled, if that ends up being done here.

src/main/scala/io/citrine/lolo/learners/ExtraRandomTrees.scala
```diff
@@ -0,0 +1,123 @@
...
+                  case x: String =>
+                    println(s"Unrecognized subsetStrategy ${x}; using auto")
+                    trainingData.head._1.size
+                }
+              case x: Int =>
+                x
+              case x: Double =>
+                (trainingData.head._1.size * x).toInt
+            }
+            val DTLearner = RegressionTreeLearner(
```
maxhutch, line 105: This should be a classification tree learner.  The fact that its passing tests suggests that this code path is untested.

src/main/scala/io/citrine/lolo/bags/Bagger.scala
```diff
@@ -15,14 +15,19 @@ import scala.reflect._
...
+  * @param uncertaintyCalibration whether to enable empirical uncertainty calibration
+  * @param disableBootstrap whether to disable bootstrap (useful when `method` implements its own randomization)
+  * */
 case class Bagger(
                    method: Learner,
                    numBags: Int = -1,
                    useJackknife: Boolean = true,
                    biasLearner: Option[Learner] = None,
                    uncertaintyCalibration: Boolean = false,
+                   disableBootstrap: Boolean = false,
```
maxhutch, line 30: I assume you added this because the standard behavior for ExtraRandomTrees is to not bootstrap.  However, this setting isn't actually used in the ExtraRandomTrees learner.

src/main/scala/io/citrine/lolo/bags/Bagger.scala
```diff
@@ -68,18 +73,22 @@ case class Bagger(
...
-      Iterator.continually {
-        Vector.fill(trainingData.size)(dist.draw())
-      }.filter(_.count(_ > 0) >= Bagger.minimumNonzeroWeightSize).take(actualBags).toVector
-    }.filter{nMat =>
-      lazy val noAlwaysPresentTrainingData = nMat.transpose.forall{vec => vec.contains(0)}
-      // Make sure that at least one learner is missing each training point
-      // This prevents a divide-by-zero error in the jackknife-after-bootstrap calcualtion
-      !useJackknife || noAlwaysPresentTrainingData
-    }.next()
+    val Nib: Vector[Vector[Int]] = if (disableBootstrap) {
```
maxhutch, line 76: I don't think this codepath is tested.  Its likely to cause some problems in the calculations below (e.g. `oobErrors` being empty) that can be handled.

src/main/scala/io/citrine/lolo/learners/ExtraRandomTrees.scala
gregor-robinson: Oh, right, that uses OOB errors.

src/main/scala/io/citrine/lolo/learners/ExtraRandomTrees.scala
```diff
@@ -0,0 +1,125 @@
...
+  * @param minLeafInstances minimum number of instances per leave in each tree
+  * @param maxDepth       maximum depth of each tree in the forest (default: unlimited)
+  * @param uncertaintyCalibration whether to empirically recalibrate the predicted uncertainties (default: false)
+  * @param disableBootstrap whether to disable bootstrap (default: true)
+  * @param randomlyRotateFeatures whether to randomly rotate real features for each tree in the forest (default: false)
+  * @param rng            random number generator to use
+  */
+case class ExtraRandomTrees(
+                             numTrees: Int = -1,
+                             useJackknife: Boolean = true,
```
maxhutch, line 35: I think we might want the default for this to be `false`; see my comment lower.

src/test/scala/io/citrine/lolo/learners/ExtraRandomTreesTest.scala
```diff
@@ -0,0 +1,177 @@
...
+      TestUtils.generateTrainingData(1024, 12, noise = 0.1, function = Friedman.friedmanSilverman, seed = rng.nextLong()),
+      inputBins = Seq((0, 8))
+    )
+
+    Seq(true, false).foreach { randomlyRotateFeatures =>
+      val RFMeta = ExtraRandomTrees(randomlyRotateFeatures = randomlyRotateFeatures, rng = rng)
+        .train(trainingData)
+      val RF = RFMeta.getModel()
+
+      assert(RFMeta.getLoss().get < 1.0, "Loss of bagger is larger than expected")
```
maxhutch, line 30: This test fails consistently for me on my local system.

src/test/scala/io/citrine/lolo/learners/ExtraRandomTreesTest.scala
```diff
@@ -0,0 +1,177 @@
...
+
+      val results = RF.transform(trainingData.map(_._1))
+      // val means = results.getExpected()
+      val sigma: Seq[Double] = results.getUncertainty().get.asInstanceOf[Seq[Double]]
+      assert(sigma.forall(_ >= 0.0))
```
maxhutch, line 35: This vector round-off level small for me: values are < 1e-12.  If bootstrapping is turned off, I'm pretty sure the jackknife method will fail or at least do something funny.

gregor-robinson:
Good call on the jackknife.

The `ExtraRandomTreesTest.testRegression` failure was clearly my fault, but I can't explain the other failures -- `StandardizerTest.testStandardLinear` for example should be perfectly deterministic.

maxhutch:
> The `ExtraRandomTreesTest.testRegression` failure was clearly my fault, but I can't explain the other failures -- `StandardizerTest.testStandardLinear` for example should be perfectly deterministic (it might, but shouldn't, fail when linked against different versions of the RNG libraries, but then it should fail reliably).

I'm getting errors on the master branch too, so let's call that "out of scope" and I'll dig into it a bit on my own.

PR #223, opened 2020-04-01
Make tests less flaky.

gregor-robinson:
There are probably a few places where tests still aren't passing through an rng parameter, but this should *dramatically* reduce the incidence of false failures, and this exposes an interface that Mithril can use to decrease its false failure rate as well.

Most of these edits are non-breaking -- I think the breaking changes are limited to the following:
* `Merit` is now a case class so that objects can have unique state (`UncertaintyCorrelation` draws random numbers)
* `StatisticalValidation` is also now a case class for the same reason.

### DIFF
--- a/src/main/scala/io/citrine/lolo/learners/RandomForest.scala
+++ b/src/main/scala/io/citrine/lolo/learners/RandomForest.scala
@@ -1,11 +1,15 @@
 package io.citrine.lolo.learners
 
+import breeze.stats.distributions.{RandBasis, ThreadLocalRandomGenerator}
 import io.citrine.lolo.bags.Bagger
 import io.citrine.lolo.transformers.FeatureRotator
 import io.citrine.lolo.trees.classification.ClassificationTreeLearner
 import io.citrine.lolo.trees.regression.RegressionTreeLearner
 import io.citrine.lolo.trees.splits.{ClassificationSplitter, RegressionSplitter}
 import io.citrine.lolo.{Learner, TrainingResult}
+import org.apache.commons.math3.random.MersenneTwister
+
+import scala.util.Random
 
 /**
   * Standard random forest as a wrapper around bagged decision trees
@@ -22,6 +26,7 @@ import io.citrine.lolo.{Learner, TrainingResult}
   * @param uncertaintyCalibration whether to empirically recalibrate the predicted uncertainties (default: false)
   * @param randomizePivotLocation whether generate splits randomly between the data points (default: false)
   * @param randomlyRotateFeatures whether to randomly rotate real features for each tree in the forest (default: false)
+  * @param randomSeed     to use for stochastic functionality (default of Long.MaxValue sets a new random seed for each invocation)
   */
 case class RandomForest(
                          numTrees: Int = -1,
@@ -33,8 +38,16 @@ case class RandomForest(
                          maxDepth: Int = Integer.MAX_VALUE,
                          uncertaintyCalibration: Boolean = false,
                          randomizePivotLocation: Boolean = false,
-                         randomlyRotateFeatures: Boolean = false
+                         randomlyRotateFeatures: Boolean = false,
+                         randomSeed: Long = Long.MaxValue
                        ) extends Learner {
+
+  val rng: Random = if (randomSeed == Long.MaxValue) {
+    new Random()
+  } else {
+    new Random(randomSeed)
+  }
+
   /**
     * Train a random forest model
     *
@@ -46,6 +59,8 @@ case class RandomForest(
     * @return training result containing a model
     */
   override def train(trainingData: Seq[(Vector[Any], Any)], weights: Option[Seq[Double]]): TrainingResult = {
+    val breezeRandBasis: RandBasis = new RandBasis(new ThreadLocalRandomGenerator(new MersenneTwister(rng.nextLong())))
+
     trainingData.head._2 match {
       case _: Double =>
         val numFeatures: Int = subsetStrategy match {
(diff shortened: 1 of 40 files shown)


src/main/scala/io/citrine/lolo/validation/CrossValidation.scala
```diff
@@ -7,7 +7,7 @@ import scala.util.Random
 /**
   * Methods tha use cross-validation to calculate predicted-vs-actual data and metric estimates
   */
-object CrossValidation {
+case class CrossValidation(rng: Random = Random) {
```
maxhutch, line 10: An alternative would be to put `rng` in the method signatures, which would let this stay as an object (basically let the methods be static)..

src/main/scala/io/citrine/lolo/validation/Merit.scala
```diff
@@ -96,7 +96,7 @@ case class StandardError(rescale: Double = 1.0) extends Merit[Double] {
   * In the absence of a closed form for that coefficient, it is model empirically by drawing from N(0, x) to produce
   * an "ideal" error series from which the correlation coefficient can be estimated.
   */
-case object UncertaintyCorrelation extends Merit[Double] {
+case class UncertaintyCorrelation(rng: Random = Random) extends Merit[Double] {
```
maxhutch, line 99: Likewise, rng could be added to `evaluate`'s signature.

src/main/scala/io/citrine/lolo/learners/RandomForest.scala
```diff
@@ -68,13 +77,16 @@ case class RandomForest(
           numFeatures = numFeatures,
           minLeafInstances = minLeafInstances,
           maxDepth = maxDepth,
-          splitter = RegressionSplitter(randomizePivotLocation)
+          splitter = RegressionSplitter(randomizePivotLocation),
+          rng = rng
```
maxhutch, line 81: What happens to this random number generator when the the training of the trees in the random forest is parallelized?

src/main/scala/io/citrine/lolo/validation/CrossValidation.scala
gregor-robinson: Do you have a preference? I went this route so that the signatures would stay the same, before I realized this would still be a breaking change.

src/main/scala/io/citrine/lolo/learners/RandomForest.scala
gregor-robinson: Paraphrasing slack discussion for posterity: parallel training breaks reproducibility. For now, this can be addressed by disabling parallelism when perfect reproducibility is required. A possible improvement, probably only worth undertaking if customers start to demand reproducibility, is to implement a threaded counter-based PRNG. (AFAICT, no such implementations exist for the JVM.)

mrupp-citrine: Why would this be necessary? This approach (essentially splitting the PRNG) should work:
https://github.com/google/jax/blob/master/design_notes/prng.md

bfolie: Maybe relevant: reproducibility in distributed environments is implemented in smlb: https://github.com/CitrineInformatics/smlb/pull/15/files

src/main/scala/io/citrine/lolo/validation/CrossValidation.scala
bfolie: I don't think it's breaking to add a new (optional) parameter to the `CrossValidation.kFoldPvA` method. And it feels less disruptive than changing an object to a case class.

src/main/scala/io/citrine/lolo/validation/Merit.scala
bfolie: Was this changed to a class to  allow for `rng` or to allow for Merits to have parameters (such as including a coverage level in the standard confidence)?

src/main/scala/io/citrine/lolo/learners/RandomForest.scala
gregor-robinson: Yep! Implementing a split approach is exactly what I meant. The main advantage of counter-based PRNGs is the ability to split them into an arbitrary number of streams that are both statistically and computationally independent. Linear PRNGs, including `scala.util.Random` and the Mersenne Twister algorithm implemented by `np.random.RandomState`, do not have this property. They are designed only to produce a single stream of statistically independent random numbers, whereas multiple streams initialized with different seeds are not statistically independent from each other. This is why the Jax notes cite the Claessen & Pałka paper, which presents a counter-based PRNG: https://dl.acm.org/doi/10.1145/2578854.2503784

But there is good news! I was being too narrow in my search terms (looking for specific algorithms) and now I've found this: https://docs.oracle.com/javase/8/docs/api/java/util/SplittableRandom.html
It's not immediately clear what algorithm it uses, but at least it passes DieHarder. That's good enough to convince me it's suitable for us, barring unexpected shortcomings in its interface or in its computational overhead. Still, I think that's outside the scope of the purpose of this PR, which is to reduce the noise of our gating tests. We can revisit if we encounter a specific motivating case for reproducibility that clearly also benefits from parallelism.

peteroupc: Java's `SplittableRandom` currently uses SplitMix, but it's [possible that this algorithm may change in future Java versions](http://openjdk.java.net/jeps/356), namely to a different algorithm called "LXM" (but I know of no article or paper that describes LXM, as opposed to SplitMix, in more detail).

src/main/scala/io/citrine/lolo/validation/Merit.scala
gregor-robinson: This was changed to allow for rng, and I flipped the others for uniformity. That's reverted now, as `rng` is now passed in `evaluate`'s signature.

src/main/scala/io/citrine/lolo/validation/CrossValidation.scala
gregor-robinson: Fair enough. Reverted.

gregor-robinson:
The goal of this work was to decrease the false-negative rate of our tests. When parallel execution is disabled, this PR also provides some reproducibility by providing random number generator seeds. However, perfect reproducibility is not well-tested. When parallel execution is enabled, the results will generally not be reproducible. However, they appear good enough to substantially reduce the incidence of false-negative tests. Should we require precise reproducibility in the future, incremental changes atop this PR will make this approach compatible with a splittable random number generator such as `java.util.SplittableRandom`.

PR #226, opened 2020-04-08
Bump version in README.

gregor-robinson:
Final edit before tagging.


maxhutch:
Heads up: this will fail the deploy parts of the job since the actual version isn't being bumped.  That's OK and expected.

PR #230, opened 2021-01-01
Switch to sbt for builds

maxhutch:
This uses sbt and the [`sbt-ci-release`](https://github.com/olafurpg/sbt-ci-release) plugin to handle builds and deploys with sbt.  This can't be fully tested until the code is merged.


latture (approved):
This looks fine to me, but someone with more travis/sbt experience might want to review these changes also.

bfolie (commented):
I don't know if I have more sbt/travis experience, but it looks fine to me as well

.travis.yml
```diff
@@ -1,49 +1,67 @@
...
-    language: python
-    python:
-    - 3.6
+    before_install:
+      - - wget http://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh
+        - chmod +x miniconda.sh
+        - ./miniconda.sh -b
+        - export PATH=/home/travis/miniconda3/bin:$PATH
+        - conda update --yes conda
+        - conda create --yes -n condaenv python=3.6
```
bfolie, line 35: Do we want to test more python versions? They're up to 3.9 now, and I don't have a good sense of what may work in 3.6 but not in newer versions.

maxhutch: We might, but this is parity with what we had, so I'd do that in another PR.  I think we'll have to do it somewhat more manually since travis doesn't have multi-language support and I could only get the sbt steps working with language: scala.

PR #232, opened 2021-01-06
Enable crossScalaVersions and make compatible with 2.12/2.13

maxhutch:
sbt provides convenient tools for cross-testing and cross-deploying
multiple scala versions.  This change enables those.  Two changes
were needed to make the code 2.12/2.13 compatible:
 - Several library versions had to be updated so both 2.12 and 2.13
   were available.  This forced a small number of code changes
   (to remove deprecated usages in breeze)
 - Parallel collections are a separate package in 2.13 and .par is
   hard to use in a compatible way, so some parallel collections
   were constructed explicitly


maxhutch:
Note the test on [`2.12.4` here](https://travis-ci.com/github/CitrineInformatics/lolo/jobs/469221805#L251) and [`2.13.4` here](https://travis-ci.com/github/CitrineInformatics/lolo/jobs/469221805#L651)

andyczerwonka (approved):
LGTM

PR #242, opened 2021-06-30
Remove sbt-launch-repo, which was a hack and isn't standard

maxhutch:
We added sbt-launch-repo to deal with a sbt hosting migration.
Now that's over, and we shouldn't need it anymore, which is good
because it is not part of standard SBT


latture (approved):

latture:
I assume we don't need a version bump for these changes?

PR #246, opened 2021-11-02
Pla 8537/implement several correlations

bfolie:
This PR implements several ways of calculating the uncertainty correlation for a multitask bagged model. Assuming the two labels are distinct and both correspond to real values, then we have the following options:
* always predict rho = 0 (this is the trivial method)
* calculate rho using the correlation over training data
* calculate rho using the bootstrap predictions
* calculate rho from the infinitesimal jackknife & jackknife-after-bootstrap covariance, divided by the square root of the associated variances

The last method is implemented in two different ways. One is more explicit and uses loops. It can be easily compared to the published method in Wager et. al. (2014). The other is more performant and uses matrix arithmetic. The two are tested against each other for consistency.

### DIFF
--- a/src/main/scala/io/citrine/lolo/bags/BaggedResult.scala
+++ b/src/main/scala/io/citrine/lolo/bags/BaggedResult.scala
@@ -1,7 +1,9 @@
 package io.citrine.lolo.bags
 
 import breeze.linalg.{DenseMatrix, DenseVector, norm}
-import io.citrine.lolo.{ParallelModelsPredictionResult, MultiTaskModelPredictionResult, PredictionResult, RegressionResult}
+import io.citrine.lolo.bags.CorrelationMethods.{Bootstrap, CorrelationMethod, FromTraining, Jackknife, Trivial}
+import io.citrine.lolo.stats.utils
+import io.citrine.lolo.{MultiTaskModelPredictionResult, ParallelModelsPredictionResult, PredictionResult, RegressionResult}
 import io.citrine.lolo.util.Async
 import org.slf4j.{Logger, LoggerFactory}
 
@@ -126,7 +128,6 @@ case class SinglePredictionBaggedResult(
     val trainingContributions = nMat.indices.toVector.map { idx =>
       // Pull the vector of the number of times this instance was used to train each tree
       val vecN = nMat(idx).toArray
-      val nTot = vecN.sum
 
       // Loop over the trees, computing the covariance for the IJ estimate and the predicted value of
       // the out-of-bag trees for the J(ackknife) estimate
@@ -135,7 +136,7 @@ case class SinglePredictionBaggedResult(
       var tNot: Double = 0.0
       var tNotCount: Int = 0
       vecN.indices.foreach { jdx =>
-        cov = cov + (vecN(jdx) - nTot) * (treePredictions(jdx) - expected)
+        cov = cov + (vecN(jdx) - 1) * (treePredictions(jdx) - expected)
 
         if (vecN(jdx) == 0) {
           tNot = tNot + treePredictions(jdx)
@@ -407,13 +408,47 @@ case class MultiPredictionBaggedResult(
   }
 }
 
+/** Enumerates the ways in which uncertainty correlation can be calculated. */
+object CorrelationMethods extends Enumeration {
+  type CorrelationMethod = Value
+
+  val Trivial, FromTraining, Bootstrap, Jackknife = Value
+}
+
 /**
   * Container with model-wise predictions for each label and the machinery to compute (co)variance.
   *
   * @param baggedPredictions  bagged prediction results for each label
   * @param realLabels         a boolean sequence indicating which labels are real-valued
+  * @param NibIn              the sampling matrix as (# bags) x (# training)
   */
-case class MultiTaskBaggedResult(baggedPredictions: Seq[BaggedResult[Any]], realLabels: Seq[Boolean]) extends BaggedResult[Seq[Any]] with MultiTaskModelPredictionResult {
+case class MultiTaskBaggedResult(
+                                  baggedPredictions: Seq[BaggedResult[Any]],
+                                  realLabels: Seq[Boolean],
+                                  NibIn: Vector[Vector[Int]],
+                                  trainingLabels: Seq[Seq[Any]],
+                                  trainingWeights: Seq[Double]
+                                ) extends BaggedResult[Seq[Any]] with MultiTaskModelPredictionResult {
+
+  lazy val Nib: Vector[Vector[Int]] = NibIn.transpose.map(_.map(_ - 1))
+
+  /* This matrix is used to compute the jackknife covariance */
+  lazy val NibJMat = new DenseMatrix[Double](Nib.head.size, Nib.size,
+    Nib.flatMap { v =>
+      val itot = 1.0 / v.size
+      val icount = 1.0 / v.count(_ == -1.0)
+      v.map(n => if (n == -1) icount - itot else -itot)
+    }.toArray
+  )
+
+  /* This matrix is used to compute the IJ covariance */
+  lazy val NibIJMat = new DenseMatrix[Double](Nib.head.size, Nib.size,
+    Nib.flatMap { v =>
+      val itot = 1.0 / v.size
+      val vtot = v.sum.toDouble / (v.size * v.size)
+      v.map(n => n * itot - vtot)
+    }.toArray
+  )
 
   override def numPredictions: Int = baggedPredictions.head.numPredictions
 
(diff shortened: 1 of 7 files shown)


src/main/scala/io/citrine/lolo/bags/BaggedResult.scala
```diff
@@ -135,7 +136,7 @@ case class SinglePredictionBaggedResult(
       var tNot: Double = 0.0
       var tNotCount: Int = 0
       vecN.indices.foreach { jdx =>
-        cov = cov + (vecN(jdx) - nTot) * (treePredictions(jdx) - expected)
+        cov = cov + (vecN(jdx) - 1) * (treePredictions(jdx) - expected)
```
bfolie, line 139: This is a bug, but it has no effect on the calculation because `\sum_b (treePredictions(b) - expected) = 0`

src/main/scala/io/citrine/lolo/bags/BaggedResult.scala
```diff
@@ -407,13 +408,47 @@ case class MultiPredictionBaggedResult(
...
   * @param realLabels         a boolean sequence indicating which labels are real-valued
+  * @param NibIn              the sampling matrix as (# bags) x (# training)
   */
-case class MultiTaskBaggedResult(baggedPredictions: Seq[BaggedResult[Any]], realLabels: Seq[Boolean]) extends BaggedResult[Seq[Any]] with MultiTaskModelPredictionResult {
+case class MultiTaskBaggedResult(
+                                  baggedPredictions: Seq[BaggedResult[Any]],
+                                  realLabels: Seq[Boolean],
+                                  NibIn: Vector[Vector[Int]],
+                                  trainingLabels: Seq[Seq[Any]],
+                                  trainingWeights: Seq[Double]
```
bfolie, line 430: `trainingLabels` and `trainingWeights` are being piped through so that the correlation over training data can be computed, but that method does not work very well and these arguments will almost certainly be removed by the time this code is released.

src/main/scala/io/citrine/lolo/bags/MultiTaskBagger.scala
```diff
@@ -105,10 +107,12 @@ class MultiTaskBaggedTrainingResult(
                                      trainingData: Seq[(Vector[Any], Seq[Any])],
                                      useJackknife: Boolean,
                                      biasModels: Seq[Option[Model[PredictionResult[Double]]]],
-                                     rescaleRatios: Seq[Double]
+                                     rescaleRatios: Seq[Double],
+                                     trainingLabels: Seq[Seq[Any]],
+                                     trainingWeights: Seq[Double]
```
bfolie, line 112: `trainingLabels` and `trainingWeights` are being piped through so that the correlation over training data can be computed, but that method does not work very well and these arguments will almost certainly be removed by the time this code is released.

src/main/scala/io/citrine/lolo/bags/MultiTaskBagger.scala
```diff
@@ -141,7 +145,9 @@ class MultiTaskBaggedModel(
                             models: ParSeq[MultiTaskModel],
                             Nib: Vector[Vector[Int]],
                             useJackknife: Boolean,
-                            biasModels: Seq[Option[Model[PredictionResult[Double]]]]
+                            biasModels: Seq[Option[Model[PredictionResult[Double]]]],
+                            trainingLabels: Seq[Seq[Any]],
+                            trainingWeights: Seq[Double]
```
bfolie, line 150: `trainingLabels` and `trainingWeights` are being piped through so that the correlation over training data can be computed, but that method does not work very well and these arguments will almost certainly be removed by the time this code is released.

src/test/scala/io/citrine/lolo/bags/MultiTaskBaggerTest.scala
```diff
@@ -166,49 +167,92 @@ class MultiTaskBaggerTest {
...
-  /**
-    * Test that multi-task with dense labels works, and remembers all its inputs
-    */
+
+  /** Test that a multi-task bagged model properly stores and transposes individual trees, and remembers labels. */
   @Test
-  def testMixed(): Unit = {
-    /* Setup some data */
-    val raw: Seq[(Vector[Double], Double)] = TestUtils.generateTrainingData(256, 12, noise = 0.1, function = Friedman.friedmanSilverman)
+  def testCombinedMultiTaskModel(): Unit = {
```
bfolie, line 173: There were two tests here doing almost the same thing, so I combined them

src/main/scala/io/citrine/lolo/bags/BaggedResult.scala
```diff
@@ -434,11 +469,155 @@ case class MultiTaskBaggedResult(baggedPredictions: Seq[BaggedResult[Any]], real
...
+          val correctionI = predMatI(::, k) - expectedI(k)
+          val correctionJ = predMatJ(::, k) - expectedJ(k)
+          val correction = correctionI.dot(correctionJ) * inverseSize2
+          0.5 * (totalCovarianceTerm(::, k) - math.E * correction)
+        }.map(_.toScalaVector())
+        // For each prediction, rectify the covariance scores to compute correlation
+        Some((scores, sigmaISeq, sigmaJSeq).zipped.map { (trainingContributions, sigmaI, sigmaJ) =>
+          BaggedResult.rectifyCorrelationScores(trainingContributions, sigmaI, sigmaJ)
+        })
+      case _: Any => None
```
sfriedowitz, line 570:
```suggestion
      case _ => None
```

src/main/scala/io/citrine/lolo/bags/BaggedResult.scala
```diff
@@ -407,13 +408,47 @@ case class MultiPredictionBaggedResult(
...
+                                  realLabels: Seq[Boolean],
+                                  NibIn: Vector[Vector[Int]],
+                                  trainingLabels: Seq[Seq[Any]],
+                                  trainingWeights: Seq[Double]
+                                ) extends BaggedResult[Seq[Any]] with MultiTaskModelPredictionResult {
+
+  lazy val Nib: Vector[Vector[Int]] = NibIn.transpose.map(_.map(_ - 1))
+
+  /* This matrix is used to compute the jackknife covariance */
+  lazy val NibJMat = new DenseMatrix[Double](Nib.head.size, Nib.size,
```
sfriedowitz, line 436: I think it would be helpful to expand the docstring on what exactly these matrices are representing (e.g. are they corresponding to direct eqns in the Wager paper)?

Also, I see that they are identical to the defns in `MultiPredictionBaggedResult`. Should the logic be off-loaded to an object method and then called once to store in the `lazy val` ?

src/main/scala/io/citrine/lolo/bags/BaggedResult.scala
```diff
@@ -434,11 +469,155 @@ case class MultiTaskBaggedResult(baggedPredictions: Seq[BaggedResult[Any]], real
...
+  }
+
+  private def getUncertaintyCorrelationJackknifeExplicit(i: Int, j: Int): Option[Seq[Double]] = {
+    val sigmaIOption = baggedPredictions(i).getUncertainty(observational = false).map(_.asInstanceOf[Seq[Double]])
+    val sigmaJOption = baggedPredictions(j).getUncertainty(observational = false).map(_.asInstanceOf[Seq[Double]])
+    (sigmaIOption, sigmaJOption) match {
+      case (Some(sigmaISeq), Some(sigmaJSeq)) =>
+        Some(Seq.tabulate(numPredictions) { predIndex => // loop over each prediction input
+          // Get the individual tree predictions for this input
+          val treePredictionsI = baggedPredictions(i).predictions.map(_.getExpected()(predIndex)).asInstanceOf[Seq[Double]]
```
sfriedowitz, line 581: If this explicit method is kept in the final version and performance is a concern, moving these lines outside the loop and then indexing them for each `predIndex` is an easy optimization.

src/main/scala/io/citrine/lolo/bags/BaggedResult.scala
```diff
@@ -434,11 +469,155 @@ case class MultiTaskBaggedResult(baggedPredictions: Seq[BaggedResult[Any]], real
...
+        // Loop over predictions, and for each one calculate the bias correction term and subtract it from each covariance term
+        // The resulting structure has size (# predictions) x (# bags)
+        val scores = baggedPredictionsI.indices.map { k =>
+          val correctionI = predMatI(::, k) - expectedI(k)
+          val correctionJ = predMatJ(::, k) - expectedJ(k)
+          val correction = correctionI.dot(correctionJ) * inverseSize2
+          0.5 * (totalCovarianceTerm(::, k) - math.E * correction)
+        }.map(_.toScalaVector())
+        // For each prediction, rectify the covariance scores to compute correlation
+        Some((scores, sigmaISeq, sigmaJSeq).zipped.map { (trainingContributions, sigmaI, sigmaJ) =>
```
sfriedowitz, line 567: I get a deprecation warning on this line: `Symbol zipped is deprecated. Use xs.lazyZip(ys).lazyZip(zs) `

src/main/scala/io/citrine/lolo/stats/utils.scala
```diff
@@ -0,0 +1,70 @@
+package io.citrine.lolo.stats
+
+import io.citrine.lolo.linear.LinearRegressionLearner
+
+import scala.util.Random
+
+object utils {
```
sfriedowitz, line 7: This is purely stylistic. But it seems like the library convention is UpperCamelCase for all objects and classes.

Should this be renamed to `StatsUtils` ?

sfriedowitz (commented):
I'm still in the process of parsing the detailed numerical bits. But here are some initial comments after my first read.

src/main/scala/io/citrine/lolo/bags/BaggedResult.scala
```diff
@@ -407,13 +408,47 @@ case class MultiPredictionBaggedResult(
   }
 }
 
+/** Enumerates the ways in which uncertainty correlation can be calculated. */
+object CorrelationMethods extends Enumeration {
```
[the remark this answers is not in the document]
bfolie: In many of those cases in Mithril, the individual values can be parameterized, so a sealed trait is necessary. In this case we want a pure enumeration, so I think we should use the built-in enumeration structure. It's also [lighter weight](https://alvinalexander.com/scala/how-to-use-scala-enums-enumeration-examples/). 

> I also propose that this enum lives in its own file in the bags module, so it is not mixed in with the definitions of all the other BaggedResult types.

Can do

src/main/scala/io/citrine/lolo/bags/BaggedResult.scala
bfolie: It's non-local, so unlikely to work in any but the most contrived examples

src/main/scala/io/citrine/lolo/bags/BaggedResult.scala
```diff
@@ -434,11 +469,155 @@ case class MultiTaskBaggedResult(baggedPredictions: Seq[BaggedResult[Any]], real
...
+    * Calculate the uncertainty correlation using one of a variety of methods.
+    * This method ensures that the two labels are distinct and both real-valued, and then dispatches the calculation
+    * to the specified method.
+    *
+    * @param i      index of the first label
+    * @param j      index of the second label
+    * @param method method by which to calculate correlation
+    * @return       optional sequence of correlation coefficients between specified labels for each prediction
+    */
+  def getUncertaintyCorrelationBuffet(i: Int, j: Int, method: CorrelationMethod): Option[Seq[Double]] = {
```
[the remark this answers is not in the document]
bfolie: In production, the user will not have a say in how the correlation is calculated. Note that this PR, and the others I've been working on, are all into a feature branch. I'd like to keep the eventual interface intact, and do my mucking around separately.

src/main/scala/io/citrine/lolo/bags/BaggedResult.scala
bfolie: Interesting. I'll look into it

src/main/scala/io/citrine/lolo/bags/BaggedResult.scala
bfolie: Yeah, that all makes sense

src/main/scala/io/citrine/lolo/bags/BaggedResult.scala
bfolie: `zipped` was deprecated in favor of `lazyZip` in Scala 2.13. Since Lolo is cross-compiled in both 2.12 and 2.13, we can't use `lazyZip` right now.

sfriedowitz: Interesting. nice munging to figure that out. I guess `zipped` lives for now.

src/main/scala/io/citrine/lolo/bags/BaggedResult.scala
```diff
@@ -507,4 +697,29 @@ object BaggedResult {
...
+    * @param sigmaX           uncertainty in first label
+    * @param sigmaY           uncertainty in second label
+    * @return                 correlation coefficient, guaranteed to be between -1.0 and 1.0
+    */
+  def rectifyCorrelationScores(covarianceScores: Vector[Double], sigmaX: Double, sigmaY: Double): Double = {
+    require(sigmaX >= 0.0 && sigmaY >= 0.0)
+    if (sigmaX == 0 || sigmaY == 0) return 0.0
+    val rho = covarianceScores.sum / (sigmaX * sigmaY)
+    // If the calculated rho is too large or too small, then the calculation is noisy and we set rho to 0.0
+    // TODO (PLA-8597): figure out a better way to rectify covariance noise
```
bfolie, line 717: I have thoughts on how to do this, but want to cover it in a separate PR in which I will benchmark the effect of the rectification strategy

src/main/scala/io/citrine/lolo/bags/BaggedResult.scala
```diff
@@ -434,11 +437,159 @@ case class MultiTaskBaggedResult(baggedPredictions: Seq[BaggedResult[Any]], real
...
+        case Bootstrap => getUncertaintyCorrelationBootstrap(i, j)
+        case Jackknife => getUncertaintyCorrelationJackknife(i, j)
+        case JackknifeExplicit => getUncertaintyCorrelationJackknifeExplicit(i, j)
+      }
+      case _: Any => None
+    }
+  }
+
+  /** Uncertainty correlation is always 0.0 */
+  private def getUncertaintyCorrelationTrivial(i: Int, j: Int): Option[Seq[Double]] = {
```
gregor-robinson, line 468: Should this be 1 on the diagonal?

src/main/scala/io/citrine/lolo/bags/BaggedResult.scala
```diff
@@ -507,4 +697,29 @@ object BaggedResult {
     }
     scores.map(Math.max(floor, _))
   } ensuring (vec => vec.forall(_ >= 0.0))
+
+  /**
+    * Calculate the correlation coefficient given a sequence of covariance scores for each training point.
+    * In theory, the covariance is the sum of the individual covariance scores. But because each covariance score
+    * is noisy, the resulting sum can be larger in absolute value than `sigmaX * sigmaY`, which would translate into
```
gregor-robinson, line 704: This comment motivates the purpose of the method, but it would be helpful to add a sentence saying what the consequence is on the resulting sum.

src/main/scala/io/citrine/lolo/stats/StatsUtils.scala
```diff
@@ -0,0 +1,70 @@
...
+
+import io.citrine.lolo.linear.LinearRegressionLearner
+
+import scala.util.Random
+
+object StatsUtils {
+
+  /** Compute the mean of a (weighted) vector, X */
+  def mean(X: Seq[Double], weights: Option[Seq[Double]] = None): Double = {
+    val actualWeights = weights.getOrElse(Seq.fill(X.length)(1.0))
```
gregor-robinson, line 11: When there are no weights, how about returning the unweighted mean instead of allocating another array and doing more complicated arithmetic when we don't have to? I'm open to this being dismissed as an unnecessary performance optimization, but it's also just a one liner here and in the immediately subsequent methods.

src/main/scala/io/citrine/lolo/stats/StatsUtils.scala
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  /** Compute the (weighted) covariance between two vectors, X and Y, of the same length. */
+  def covariance(X: Seq[Double], Y: Seq[Double], weights: Option[Seq[Double]] = None): Double = {
+    require(X.length == Y.length, s"Cannot compute covariance between sequences of different lengths (lengths are ${X.length} and ${Y.length}).")
+    val actualWeights = weights.getOrElse(Seq.fill(X.length)(1.0))
+    val totalWeight = actualWeights.sum
+    val muX = mean(X, Some(actualWeights))
+    val muY = mean(Y, Some(actualWeights))
+    (X, Y, actualWeights).zipped.map { case (x, y, w) => (x - muX) * (y - muY) * w }.sum / totalWeight
```
gregor-robinson, line 34: Is it intentional not to mirror the bias correction pattern used in `variance`?

src/main/scala/io/citrine/lolo/bags/BaggedResult.scala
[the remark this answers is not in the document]
gregor-robinson: This name makes me hungry.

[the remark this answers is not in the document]
bfolie: all-you-can-eat-uncertainty

src/main/scala/io/citrine/lolo/bags/BaggedResult.scala
```diff
@@ -260,8 +263,8 @@ case class MultiPredictionBaggedResult(
 
   override def numPredictions: Int = expectedMatrix.length
 
-  /* Subtract off 1 to make correlations easier; transpose to be prediction-wise */
-  lazy val Nib: Vector[Vector[Int]] = NibIn.transpose.map(_.map(_ - 1))
+  /* transpose to be training-wise */
```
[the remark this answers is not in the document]
bfolie: I'll change it to "transpose to be (N_training) x (N_models)"

src/main/scala/io/citrine/lolo/bags/BaggedResult.scala
bfolie: That's taken care of by the first case of `getUncertaintyCorrelationBuffet`

src/main/scala/io/citrine/lolo/stats/StatsUtils.scala
bfolie: only because a use case for it didn't come up

src/main/scala/io/citrine/lolo/stats/StatsUtils.scala
```diff
@@ -0,0 +1,70 @@
...
+    * The procedure is to generate new data randomly, calculate the residuals of a linear regression,
+    * and then create a suitable linear combination of X and the residuals.
+    * Source: https://stats.stackexchange.com/a/313138
+    *
+    * @param X    sequence of values
+    * @param rho  desired Pearson correlation coefficient
+    * @param rng  random number generator
+    * @return     sequence of values that have desired correlation with X
+    */
+  def makeLinearCorrelatedData(X: Seq[Double], rho: Double, rng: Random = new Random()): Seq[Double] = {
```
sfriedowitz, line 56: Is this function going to be used in part of the main package? Or just as a test function? It seems like a test utility to me and could be moved to an appropriate location in the test suite.

Since this is a feature branch merge, it's not high priority for now.

sfriedowitz (approved):
This looks good for now, with my one comment if you'd like to address now/later.

src/main/scala/io/citrine/lolo/stats/StatsUtils.scala
bfolie: Is it just a one-liner? I thought it would involve an ugly if-else block for each method, which is why I decided to take the performance hit in exchange for simpler-looking methods.

src/main/scala/io/citrine/lolo/stats/StatsUtils.scala
bfolie: You're right that this is unlikely to be useful outside of testing/benchmarking. I'll move it as part of the benchmarking PR, which I'm working on now.

bfolie:
I just pushed some changes to comments and requested re-review. But then I remembered that this is a PR into a feature branch, so re-review is not actually necessary

PR #247, opened 2021-11-03
Pla 8566/multitask predicted-vs-actual

bfolie:
Implement `getPredictedVsActual` for multitask bagged model, and in doing so allow for an implementation of `getLoss`. Because we expect to use multitask bagged models inside of a standardizer model, we also implement `getPredictedVsActual` for standardizers.

### DIFF
--- a/src/main/scala/io/citrine/lolo/TrainingResult.scala
+++ b/src/main/scala/io/citrine/lolo/TrainingResult.scala
@@ -39,5 +39,5 @@ trait MultiTaskTrainingResult extends TrainingResult {
 
   def getModels(): Seq[Model[PredictionResult[Any]]]
 
-  override def getPredictedVsActual(): Option[Seq[(Vector[Any], Seq[Any], Seq[Any])]] = None
+  override def getPredictedVsActual(): Option[Seq[(Vector[Any], Seq[Option[Any]], Seq[Option[Any]])]] = None
 }
--- a/src/main/scala/io/citrine/lolo/bags/Bagger.scala
+++ b/src/main/scala/io/citrine/lolo/bags/Bagger.scala
@@ -2,7 +2,7 @@ package io.citrine.lolo.bags
 
 import breeze.linalg.DenseMatrix
 import breeze.stats.distributions.{Poisson, Rand, RandBasis}
-import io.citrine.lolo.stats.metrics.ClassificationMetrics
+import io.citrine.lolo.stats.metrics.{ClassificationMetrics, RegressionMetrics}
 import io.citrine.lolo.util.{Async, InterruptibleExecutionContext}
 import io.citrine.lolo.{Learner, Model, PredictionResult, RegressionResult, TrainingResult}
 
@@ -210,10 +210,8 @@ class BaggedTrainingResult[+T : ClassTag](
   }
 
   lazy val loss: Double = rep match {
-    case _: Double => Math.sqrt(predictedVsActual.map(d => Math.pow(d._2.asInstanceOf[Double] - d._3.asInstanceOf[Double], 2)).sum / predictedVsActual.size)
-    case _: Any =>
-      val f1 = ClassificationMetrics.f1scores(predictedVsActual)
-      if (f1 > 0.0) 1.0 / f1 - 1.0 else Double.MaxValue
+    case _: Double => RegressionMetrics.RMSE(predictedVsActual.asInstanceOf[Seq[(Vector[Any], Double, Double)]])
+    case _: Any => ClassificationMetrics.loss(predictedVsActual)
   }
 
   /**
--- a/src/main/scala/io/citrine/lolo/bags/MultiTaskBagger.scala
+++ b/src/main/scala/io/citrine/lolo/bags/MultiTaskBagger.scala
@@ -110,6 +110,28 @@ class MultiTaskBaggedTrainingResult(
 
   lazy val model = new MultiTaskBaggedModel(models, Nib, useJackknife, biasModels)
 
+  val predictedVsActual = trainingData.zip(Nib.transpose).flatMap { case ((features, labels), nb) =>
+    // Bagged models that were not trained on this input
+    val oob = models.zip(nb).filter(_._2 == 0).map(_._1)
+    if (oob.isEmpty) {
+      Seq()
+    } else {
+      // "Average" the predictions on each label over the out-of-bag models
+      val oobPredictions = oob.map(_.transform(Seq(features)).getExpected().head)
+      val predicted = oobPredictions.toVector.transpose.zipWithIndex.map {
+        case (predictions, labelIndex) if models.head.getRealLabels(labelIndex) =>
+          predictions.asInstanceOf[Seq[Double]].sum / predictions.size
+        case (predictions, _) => predictions.groupBy(identity).maxBy(_._2.size)._1
+      }
+      // Remove predictions if the label was not specified
+      val (optionLabels, optionPredicted) = labels.zip(predicted).map {
+        case (l, _) if l == null || (l.isInstanceOf[Double] && l.asInstanceOf[Double].isNaN) => (None, None)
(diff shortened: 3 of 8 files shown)


src/main/scala/io/citrine/lolo/TrainingResult.scala
```diff
@@ -39,5 +39,5 @@ trait MultiTaskTrainingResult extends TrainingResult {
 
   def getModels(): Seq[Model[PredictionResult[Any]]]
 
-  override def getPredictedVsActual(): Option[Seq[(Vector[Any], Seq[Any], Seq[Any])]] = None
+  override def getPredictedVsActual(): Option[Seq[(Vector[Any], Seq[Option[Any]], Seq[Option[Any]])]] = None
```
bfolie, line 42: Because not every training datum needs to have a value for every label, this interface needs to expose `Seq[Option[Any]]`

src/main/scala/io/citrine/lolo/bags/Bagger.scala
```diff
@@ -210,10 +210,8 @@ class BaggedTrainingResult[+T : ClassTag](
   }
 
   lazy val loss: Double = rep match {
-    case _: Double => Math.sqrt(predictedVsActual.map(d => Math.pow(d._2.asInstanceOf[Double] - d._3.asInstanceOf[Double], 2)).sum / predictedVsActual.size)
-    case _: Any =>
-      val f1 = ClassificationMetrics.f1scores(predictedVsActual)
-      if (f1 > 0.0) 1.0 / f1 - 1.0 else Double.MaxValue
+    case _: Double => RegressionMetrics.RMSE(predictedVsActual.asInstanceOf[Seq[(Vector[Any], Double, Double)]])
+    case _: Any => ClassificationMetrics.loss(predictedVsActual)
```
bfolie, line 214: The same logic is going to be used for `MultiTaskBagger`, so write it once

src/main/scala/io/citrine/lolo/transformers/Standardizer.scala
```diff
@@ -83,6 +92,19 @@ class StandardizerTrainingResult(
...
+
+  override def getPredictedVsActual(): Option[Seq[(Vector[Any], Any, Any)]] = {
+    baseTrainingResult.getPredictedVsActual() match {
+      case None => None
+      case Some(predictedVsActual) => Some(
+        (
+          Standardizer.invertStandardization(predictedVsActual.map(_._1), inputTrans),
+          Standardizer.invertStandardization(predictedVsActual.map(_._2), outputTrans),
+          Standardizer.invertStandardization(predictedVsActual.map(_._3), outputTrans),
+        ).zipped.toSeq
```
sfriedowitz, line 104: I get the same deprecation warning here for `.zipped`

sfriedowitz (commented):
This looks pretty good overall. A few comments of questions/thoughts.

src/main/scala/io/citrine/lolo/bags/MultiTaskBagger.scala
```diff
@@ -110,23 +111,67 @@ class MultiTaskBaggedTrainingResult(
...
+      // "Average" the predictions on each label over the out-of-bag models
+      val oobPredictions = oob.map(_.transform(Seq(features)).getExpected().head)
+      val predicted = oobPredictions.toVector.transpose.zipWithIndex.map {
+        case (predictions, labelIndex) if models.head.getRealLabels(labelIndex) =>
+          predictions.asInstanceOf[Seq[Double]].sum / predictions.size
+        case (predictions, _) => predictions.groupBy(identity).maxBy(_._2.size)._1
+      }
+      // Remove predictions for which the label was not specified
+      val (optionLabels, optionPredicted) = labels.zip(predicted).map {
+        case (l, _) if l == null || (l.isInstanceOf[Double] && l.asInstanceOf[Double].isNaN) => (None, None)
```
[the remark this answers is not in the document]
bfolie: I think that `null` gets used when the label is missing

src/main/scala/io/citrine/lolo/transformers/Standardizer.scala
```diff
@@ -2,6 +2,15 @@ package io.citrine.lolo.transformers
 
 import io.citrine.lolo._
 
+case class Standardization(shift: Double, scale: Double) {
+  require(scale > 0 && scale < Double.PositiveInfinity)
```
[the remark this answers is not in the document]
bfolie: good point, done

src/main/scala/io/citrine/lolo/transformers/Standardizer.scala
bfolie: We can't use `lazyZip` because we compile with Scala 2.12 in addition to 2.13

gregor-robinson: Not really related to this PR but... do you know how long we can expect to maintain that?

src/main/scala/io/citrine/lolo/bags/MultiTaskBagger.scala
```diff
@@ -110,23 +111,68 @@ class MultiTaskBaggedTrainingResult(
 
   lazy val model = new MultiTaskBaggedModel(models, Nib, useJackknife, biasModels)
 
+  lazy val predictedVsActual: Seq[(Vector[Any], Seq[Option[Any]], Seq[Option[Any]])] =
```
gregor-robinson, line 114: Slinging tuples is hard to read. Going forward, would be nice to use something more specialized so that we can write e.g. `_.actual` instead of `_._3`. For now, let's shore this up a little by adding a comment here that documents what each item in the tuple means. (It's somewhat self-documenting at the end of this var assignment block, but some prose before line 114 would be choice.)

src/main/scala/io/citrine/lolo/transformers/Standardizer.scala
```diff
@@ -170,7 +203,7 @@ class StandardizerPrediction[T](
     */
   override def getUncertainty(includeNoise: Boolean = true): Option[Seq[Any]] = {
     baseResult.getUncertainty(includeNoise) match {
-      case Some(x) if outputTrans.isDefined => Some(x.map(_.asInstanceOf[Double] * rescale))
+      case Some(x) if outputTrans.isDefined => Some(x.map(_.asInstanceOf[Double] * outputRescale))
```
gregor-robinson, line 206: I wonder if there's a straightforward way to do the type resolution on `x` instead of within the `map`.

src/main/scala/io/citrine/lolo/transformers/Standardizer.scala
bfolie: That's a good question. I think that we made Lolo cross-compatible to facilitate the Mithril migration. Now that that's done, perhaps we can bump Lolo to 2.13.
On the other hand, since it is a public library, there might be value in maintaining backwards compatibility for a while.

src/main/scala/io/citrine/lolo/transformers/Standardizer.scala
bfolie: I think it's not possible because of type erasure

sfriedowitz: Yeah, you will not be able to match on `Seq[Double]` becasue of erasure. You could with a `ClassTag` parameter I believe, but that would require changing the interface for all `getUncertainty` methods.

sfriedowitz (approved):
This LGTM @bfolie

gregor-robinson (approved):

src/main/scala/io/citrine/lolo/bags/MultiTaskBagger.scala
bfolie: I forgot to add these comments before, but I just added them. Because the merge is into a feature branch, I don't need re-review

PR #248, opened 2021-11-09
introduce factor of (n-1)/n in jackknife bias correction terms

bfolie:
The bias correction terms in Wager et. al. (2014) [1] (equations 7 and 8) are slightly incorrect. They have factors of `n` that should be `n-1`.

The infinitesimal jackknife (IJ) bias term is derived in section 4.1 of [1]. Just before equation 11, they state `Var[N_{bi}] = 1`, where `N_{bi}` is the number of times training point `i` appears in bag `b`. `N_{bi}` follows a multinomial distribution with probabilities `p_i = 1/n` and `n` draws. The variance of this quantity is `(n-1)/n`, not 1, which is why their final result is missing a factor of `(n-1)/n`. This mistake is also caught in Appendix A.2 of [2], in which the authors re-derive the IJ equations for their own edification. See the section after equation A.4, in particular.

The Jackknife after Bootstrap bias term is derived in Appendix A of [1]. The equations are, frustratingly, not numbered, but the next-to-last expression for the bias has a factor of `n-1` (it can be found after the sentence "Finally, this lets us conclude that"). But then in the final rendition, after "This lets us considerably simplify our expressions for Monte Carlo bias and variance", the factor of `n-1` has been approximated into a factor of `n` without explanation. I suspect that this was because they wanted it to be easily combine-able with the IJ bias correction.

[1] Wager, S.; Hastie, T. and Efron, B. Confidence Intervals for Random Forests: The Jackknife and the Infinitesimal Jackknife. Journal of Machine Learning Research 15 (2014) 1625-1651.
[2] Su, X.; Pena, A. T.; Liu, L. and Levine, R. A. Random Forests of Interaction Trees for Estimating Individualized Treatment Effects in Randomized Trials. Statistics in Medicine 37 (2018).

### DIFF
--- a/src/main/scala/io/citrine/lolo/bags/BaggedResult.scala
+++ b/src/main/scala/io/citrine/lolo/bags/BaggedResult.scala
@@ -115,12 +115,12 @@ case class SinglePredictionBaggedResult(
   override def getImportanceScores(): Option[Seq[Seq[Double]]] = Some(Seq(singleScores))
 
   private lazy val singleScores: Vector[Double] = {
-    // Compute the Bessel-uncorrected variance of the ensemble of predicted values,
-    // and then divide by the size of the ensemble an extra time
-    val varT = treeVariance * (treePredictions.length - 1.0) / (treePredictions.length * treePredictions.length)
-
     // This will be more convenient later
     val nMat = NibIn.transpose
+
+    // Compute the Bessel-uncorrected variance of the ensemble of predicted values, then multiply by (n - 1) / (n * B)
+    // We later sum over the training data, introducing a factor of n and leaving us with the expected correction term
+    val correction = treeVariance * (treePredictions.length - 1.0) * (nMat.size - 1) / (treePredictions.length * treePredictions.length * nMat.size)
 
     // Loop over each of the training instances, computing its contribution to the uncertainty
     val trainingContributions = nMat.indices.toVector.map { idx =>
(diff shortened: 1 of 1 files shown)


gregor-robinson:
If it's okay with you @bfolie, I want to hold off on this until tracing through what I suspect might be another mistake.

bfolie:
> If it's okay with you @bfolie, I want to hold off on this until tracing through what I suspect might be another mistake in Wager.

😱 

👍

bfolie:
> If it's okay with you @bfolie, I want to hold off on this until tracing through what I suspect might be another mistake in Wager.

Upon further thought, I'd like to merge this before we resolve the other issue. Since the possible mistake in Efron's work is a factor of sqrt(n/(n-1)), it won't perfectly cancel the factor introduced here. Also, the additional term would apply to both the IJ and the IJ bias correction. Hence, this PR makes the two terms correct _relative to each other_. And if there is a mistake in Efron that necessitates an additional factor, it would be nicer to make that change in a single PR that modifies both terms, rather than combining it with the factor in this change.

src/main/scala/io/citrine/lolo/bags/BaggedResult.scala
```diff
@@ -351,13 +348,13 @@ case class MultiPredictionBaggedResult(
...
-    /* Avoid division in the loop */
-    val inverseSize = 1.0 / modelPredictions.head.size
+    /* Avoid division in the loop, calculate (n - 1) / (n * B^2) */
+    val prefactor = 1.0 / Math.pow(modelPredictions.head.size, 2.0) * (Nib.size - 1) / Nib.size
 
     modelPredictions.indices.map { i =>
       Async.canStop()
       /* Compute the first order bias correction for the variance estimators */
-      val correction = Math.pow(inverseSize * norm(predMat(::, i) - meanPrediction(i)), 2)
+      val correction = prefactor * Math.pow(norm(predMat(::, i) - meanPrediction(i)), 2)
```
gregor-robinson, line 357: Insofar as avoiding division in the loop is a relevant optimization here, s/`Math.pow(norm(predMat(::, i) - meanPrediction(i)), 2)`/`norm(predMat(::, i) - meanPrediction(i)) * norm(predMat(::, i) - meanPrediction(i))`/ (where `norm(predMat(::, i) - meanPrediction(i))` is actually factored out into its own `val` preceding this line) might also be worthwhile. Unless `Math.pow` includes that optimization already. (But of course that's beyond the point of this PR.)

gregor-robinson (approved):

src/main/scala/io/citrine/lolo/bags/BaggedResult.scala
bfolie: Looking at the docstring for Java's `pow` method, which lists several special cases, it doesn't look like there's a special case for "the second argument is an integer" (there is a special case for "both arguments are integers). So there might be room for improvement in several places by writing out `x * x` instead of `math.pow(x, 2)`

PR #249, opened 2021-11-12
Pla 8639/correlation test harness

bfolie:
Create a harness that we can use to study the accuracy of correlation estimates. This code is not expected to make its way into main. Instead, it will turn into a stand-alone branch that one will be able to run to reproduce the results of the upcoming correlation study.

You can pull the branch and run `main` to verify that it makes a figure and csv. I'm mostly looking for the following things in a review:
* suggestions for other metrics that would be worth including
* suggestions for other test problems that would be worth including (see below for a discussion of the current metrics and test problems)
* a check for bugs in the metric code
* general methodological suggestions

I'm less interested in suggestions around code organization because this code won't go into production, and the primary goal is to make it easy to run multiple trials sequentially. There's some boiler plate, especially around adapting the `plotMeritScan` function so that we can switch out the independent variable. But I don't think that's an issue in this case.

For a given set of parameters, a single trial goes as follows:
1. Generate ground-truth data according to some function, y = f(x)
2. Generate a label that has some fixed linear correlation with y. There's a parameter, `trainRho`, that can vary from 0 (uncorrelated) to 1 (perfectly correlated)
3. Generate a label that has a quadratic relationship with y. This is done by setting `z = (y-\bar{y})^2` and then adding a normally distributed random variable to `z`. The scale of this random variable, called "fuzz" allows us to tune the relationship. At `fuzz=0` the two variables are strongly correlated, albeit in a non-linear way. As `fuzz->\infty`, the two variables become uncorrelated.
4. Add normally distributed sampling noise to all of the labels. This gives us the "experimentally observed" values.
5. Split the data into train and test sets
6. Train a multitask random forest on the noised training data
7. Apply the trained model to the test data
8. Apply some metric (either NLPD or standard confidence) to the predictions. Whether we compare the prediction to the ground-truth values or the noised values depends on whether `observational` equals true or false (are we judging the ability to compute a confidence interval or to compute a prediction interval)

This is repeated `numTrials` times to get a mean and standard error, and some parameter is swept to produce a graph.

I made the reproducibility as high as I could. The generation of training/test data, the application of sampling noise, the Poisson draws for creating the bags, and the rng that is passed to training are all repeatable. **But** the bags are trained in parallel and the order in which they are trained is not repeatable. Since we don't have splittable random numbers, this means that a given bag will encounter the rng in a different state, depending on the order of training.

Though not perfect, re-running a set of trials produces very similar results. It's repeatable enough that when I introduced a bug (see commit #88477d3), I was able to catch it because the results changed in a non-trivial way.

### DIFF
--- a/src/main/scala/io/citrine/lolo/validation/Merit.scala
+++ b/src/main/scala/io/citrine/lolo/validation/Merit.scala
@@ -160,7 +160,7 @@ case class NegativeLogProbabilityDensity2d(i: Int, j: Int, method: CorrelationMe
   def NLPD(pvas: Seq[PredictedVsActualTwoDimensions]): Seq[Double] = {
     pvas.map { pva =>
       val term1 = math.log(2 * math.Pi * pva.sigmaX * pva.sigmaY * math.sqrt(1 - math.pow(pva.rho, 2.0)))
-      val term2 = 1/2 * pva.mahalanobisSquared
+      val term2 = 0.5 * pva.mahalanobisSquared
       term1 + term2
     }
   }
--- a/src/main/scala/io/citrine/lolo/validation/Merit.scala
+++ b/src/main/scala/io/citrine/lolo/validation/Merit.scala
@@ -132,79 +132,65 @@ case object UncertaintyCorrelation extends Merit[Double] {
     val covar = error.zip(sigma).map { case (x, y) => (x - meanError) * (y - meanSigma) }.sum / sigma.size
     covar / Math.sqrt(varError * varSigma)
   }
+}
 
-  /**
-    * Negative Log Probability Density (NLPD) in two dimensions
-    * NLPD is calculated for each point and the median value is returned.
-    *
-    * @param i index of the first label to be compared
-    * @param j index of the second label to be compared
-    * @param method method to calculate correlation coefficient
-    * @param observational whether or not to calculate the observational uncertainty
-    */
-  case class NegativeLogProbabilityDensity2d(i: Int, j: Int, method: CorrelationMethod, observational: Boolean) extends Merit[Seq[Any]] {
-    override def evaluate(predictionResult: PredictionResult[Seq[Any]], actual: Seq[Seq[Any]], rng: Random): Double = {
-      val allPredictions = predictionResult.getExpected()
-      val predictionsI = extractComponentByIndex(allPredictions, i)
-      val predictionsJ = extractComponentByIndex(allPredictions, j)
-      val actualI = extractComponentByIndex(actual, i)
-      val actualJ = extractComponentByIndex(actual, j)
-      val errorI = predictionsI.zip(actualI).map(pa => pa._1 - pa._2)
-      val errorJ = predictionsJ.zip(actualJ).map(pa => pa._1 - pa._2)
-      val allSigma = predictionResult.asInstanceOf[MultiTaskBaggedResult].getUncertainty(observational).get
-      val sigmaI = extractComponentByIndex(allSigma, i, Some(0.0))
-      val sigmaJ = extractComponentByIndex(allSigma, j, Some(0.0))
-      val rho = predictionResult.asInstanceOf[MultiTaskBaggedResult].getUncertaintyCorrelationBuffet(i, j, method).get
-      val nlpd = NLPD(errorI, errorJ, sigmaI, sigmaJ, rho)
-      StatsUtils.median(nlpd)
-    }
-
-    /**
-      * A convenience method for slicing a two dimensional data structure
-      *
-      * @param data a sequence of sequences
-      * @param index to slice (must correspond to real-valued data)
-      * @param default default to use in case the values are Option[Double]
-      * @return
-      */
-    private def extractComponentByIndex(data: Seq[Seq[Any]], index: Int, default: Option[Double] = None): Seq[Double] = {
-      data.map { row =>
-        row(index) match {
-          case x: Double => x
-          case x: Option[Double] => x.getOrElse(default.get)
-        }
-      }
-    }
-
-    /**
-      * Bivariate negative log probability density calculation
-      * log(2 * pi * sigma_x * sigma_y * sqrt(1 - rho**2)) +
-      *   1/(2 * (1 - rho**2)) * [(x / sigma_x)**2 + (y / sigma_y)**2 - 2 * rho * x * y / (sigma_x * sigma_y)]
-      * @param dx sequence of prediction errors in first label
-      * @param dy sequence of prediction errors in second label
-      * @param sigmaX sequence of uncertainty values of first label
-      * @param sigmaY sequence of uncertainty values of second label
-      * @param rho sequence of correlation coefficients for predictions
-      * @return sequence of NLPD value for each prediction
-      */
-    def NLPD(dx: Seq[Double], dy: Seq[Double], sigmaX: Seq[Double], sigmaY: Seq[Double], rho: Seq[Double]): Seq[Double] = {
-      Seq(dx, dy, sigmaX, sigmaY, rho).transpose.map { values =>
-        val x = values.head
-        val y = values(1)
-        val sx = values(2)
-        val sy = values(3)
-        val r = values(4)
-        val r2 = math.pow(r, 2.0)
-        val term1 = math.log(2 * math.Pi * sx * sy * math.sqrt(1 - r2))
-        val normX = x / sx
-        val normY = y / sy
-        val term2 = 1/(2 * (1 - r2)) * (math.pow(normX, 2.0) + math.pow(normY, 2.0) - 2 * r * normX * normY)
-        term1 + term2
-      }
-    }
+/**
+  * Negative Log Probability Density (NLPD) in two dimensions
+  * NLPD is calculated for each point and the median value is returned.
+  *
+  * @param i index of the first label to be compared
+  * @param j index of the second label to be compared
+  * @param method method to calculate correlation coefficient
+  * @param observational whether or not to calculate the observational uncertainty
+  */
+case class NegativeLogProbabilityDensity2d(i: Int, j: Int, method: CorrelationMethod, observational: Boolean) extends Merit[Seq[Any]] {
+  override def evaluate(predictionResult: PredictionResult[Seq[Any]], actual: Seq[Seq[Any]], rng: Random): Double = {
+    val pvas = PredictedVsActualTwoDimensions.makePva2d(predictionResult, actual, i, j, method, observational)
+    val nlpd = NLPD(pvas)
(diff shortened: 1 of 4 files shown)


src/test/scala/io/citrine/lolo/bags/CorrelationStudy.scala
```diff
@@ -0,0 +1,440 @@
...
+import io.citrine.lolo.stats.functions.Friedman
+import io.citrine.lolo.trees.multitask.MultiTaskTreeLearner
+import io.citrine.lolo.validation.{Merit, NegativeLogProbabilityDensity2d, StandardConfidence2d}
+import org.apache.commons.math3.random.MersenneTwister
+import org.knowm.xchart.{BitmapEncoder, XYChart}
+import org.knowm.xchart.BitmapEncoder.BitmapFormat
+
+import java.io.{BufferedWriter, File, FileWriter}
+import scala.util.Random
+
```
bfolie, line 15: I made heavy use of sealed traits (essentially enums) to codify the different options we have for running tests and to make it easy to add new options

src/test/scala/io/citrine/lolo/bags/CorrelationStudy.scala
```diff
@@ -0,0 +1,440 @@
...
+case object NumTraining extends VariedParameter {
+  def name: String = "number of training rows"
+}
+
+object CorrelationStudy {
+
+  def main(args: Array[String]): Unit = {
+    val mainRng = new Random(52109317L)
+
+    runTrialsAndSave(
```
bfolie, line 83: This is a single example. Ultimately, this `main` method will be a series of `runTrialsAndSave` calls, one for each scan we want to perform. Each such call produces a chart and a csv

src/test/scala/io/citrine/lolo/bags/CorrelationStudy.scala
```diff
@@ -0,0 +1,440 @@
...
+    }
+    Merit.plotMeritScan(
+      parameterName = variedParameter.name,
+      parameterValues = parameterValues,
+      merits = merits,
+      rng = new Random(0L) // this is irrelevant unless we were computing Uncertainty Correlation
+    )(pvaBuilder)
+  }
+
+  def runTrial(
```
bfolie, line 299: This method trains the models and makes the predictions

sfriedowitz:
I'm just popping in here. But at a high level, this seems like it would make more sense to be a standalone (internal) repo that just depends on a fixed Lolo version for reproducing our study.

Are there pieces of the code that you need for your study that are not exposed in the public interface? What has been the standard in the past for this type of work?

bfolie:
> I'm just popping in here. But at a high level, this seems like it would make more sense to be a standalone (internal) repo that just depends on a fixed Lolo version for reproducing our study.
> 
> Are there pieces of the code that you need for your study that are not exposed in the public interface? What has been the standard in the past for this type of work?

No, I don't think there are any barriers to making this a stand-alone repo. Typically, for these kinds of small-scale studies we have generally created a branch that produces the results, and linked to that branch in the study write-up. This is mostly for simplicity.

What do you see as the primary benefits of making a stand-alone repo?

src/test/scala/io/citrine/lolo/bags/CorrelationStudy.scala
```diff
@@ -0,0 +1,441 @@
...
+          "Trivial" -> StandardConfidence2d(0, index, CorrelationMethods.Trivial, observational),
+          "Training Data" -> StandardConfidence2d(0, index, CorrelationMethods.FromTraining, observational),
+          "Bootstrap" -> StandardConfidence2d(0, index, CorrelationMethods.Bootstrap, observational),
+          "Jackknife" -> StandardConfidence2d(0, index, CorrelationMethods.Jackknife, observational)
+        )
+    }
+    // There's some unfortunate boiler plate here. plotMeritScan takes a function that goes from a Double parameter value
+    // to an iterator of what is essentially predicted-vs-actuals. This always uses the runTrial method, but for each
+    // case we need to separately pick where the parameter gets plugged in.
+    val pvaBuilder: Double => Iterator[(PredictionResult[Seq[Any]], Seq[Seq[Any]])] = variedParameter match {
```
sfriedowitz, line 215: Not the cleanest, but this is just for running your research tests. So I don't see any issue.

src/main/scala/io/citrine/lolo/validation/Merit.scala
```diff
@@ -215,3 +275,95 @@ object Merit {
...
+  * A single predicted-vs-actual point in two real-valued dimensions.
+  * The covariance matrix of the prediction has diagonals sigmaX**2 and sigmaY**2, and off-diagonals rho*sigmaX*sigmaY
+  *
+  * @param dx the error along the first dimension
+  * @param dy the error along the second dimension
+  * @param sigmaX the uncertainty along the first dimension
+  * @param sigmaY the uncertainty along the second dimension
+  * @param rho the correlation coefficient
+  */
+case class PredictedVsActualTwoDimensions(
```
sfriedowitz, line 289: It seems like this could generalize with the Mahalanobis distance to N-dimensions quite easily. Would it make sense to test >2 dimensions for some of your upcoming examples?

src/test/scala/io/citrine/lolo/bags/CorrelationStudy.scala
```diff
@@ -0,0 +1,441 @@
...
+      numTrain + numTest,
+      function.numCols,
+      noise = 0.0, // Add noise later, after computing covariate labels
+      function = function.function,
+      seed = dataGenSeed
+    )
+
+    val inputs: Seq[Vector[Double]] = fullData.map(_._1)
+    val realLabel: Seq[Double] = fullData.map(_._2)
+    val linearLabel: Seq[Double] = makeLinearCorrelatedData(realLabel, rhoTrain, linearCorrelationRng)
```
sfriedowitz, line 328: Would it be helpful to try an example with a sinusoidal map on the test function output?

I do not have any prior that this would be useful, besides the fact that it's a natural extension to a different function type.

src/main/scala/io/citrine/lolo/validation/Merit.scala
```diff
@@ -132,6 +134,64 @@ case object UncertaintyCorrelation extends Merit[Double] {
...
+
+  /**
+    * Bivariate negative log probability density calculation
+    * log(2 * pi * sigma_x * sigma_y * sqrt(1 - rho**2)) + r**2/2, where r is the Mahalanobis distance
+    * @param pvas sequence of bivariate predicted-vs-actual points
+    * @return sequence of NLPD value for each prediction
+    */
+  def NLPD(pvas: Seq[PredictedVsActualTwoDimensions]): Seq[Double] = {
+    pvas.map { pva =>
+      val term1 = math.log(2 * math.Pi * pva.sigmaX * pva.sigmaY * math.sqrt(1 - math.pow(pva.rho, 2.0)))
```
sfriedowitz, line 161: `pva.sigmaX * pva.sigmaY * math.sqrt(1 - math.pow(pva.rho, 2.0)` this is det(Cov) in two dimensions?

bfolie:
> I'm just popping in here. But at a high level, this seems like it would make more sense to be a standalone (internal) repo that just depends on a fixed Lolo version for reproducing our study.

I agree now. I'll make it a stand-alone repo so that I can continue tinkering with the study even after we've merged to Lolo's main branch. There are questions that we don't need to answer to push Lolo forward, but that would be valuable to investigate for a publication (for example, testing against a real dataset or checking convergence of the covariance estimate), and it makes sense to put those in a stand-alone repo.

So I will move this over, but it's not my immediate priority and the same code will need review in the new repo, so I could use still use a review here.

src/main/scala/io/citrine/lolo/validation/Merit.scala
```diff
@@ -132,6 +134,64 @@ case object UncertaintyCorrelation extends Merit[Double] {
...
+  * @param i index of the first label to be compared
+  * @param j index of the second label to be compared
+  * @param method method to calculate correlation coefficient
+  * @param observational whether or not to calculate the observational uncertainty
+  */
+case class NegativeLogProbabilityDensity2d(i: Int, j: Int, method: CorrelationMethod, observational: Boolean) extends Merit[Seq[Any]] {
+  override def evaluate(predictionResult: PredictionResult[Seq[Any]], actual: Seq[Seq[Any]], rng: Random): Double = {
+    val pvas = PredictedVsActualTwoDimensions.makePva2d(predictionResult, actual, i, j, method, observational)
+    val nlpd = NLPD(pvas)
+    StatsUtils.median(nlpd)
```
sfriedowitz, line 150: Would it be useful to look at max(NLPD) to try to get a sense of how bad the "worst" prediction is in terms of how far the actual falls outside the predicted uncertainty region? 

Or is this accounted for by the PVA2D metric that uses a cutoff on the Mahalaonbis distance?

sfriedowitz (approved):
I'm leaving off comments on code structure since that is not the point of this PR/review.

I left a few (possibly naive) thoughts on some of the tests you're running. I have not ran them all locally and analyzed the data myself, so I will wait for your slides to give input on that. But as far as the set up goes, I think the idea of mapping "known" correlations onto a test function makes sense.

src/main/scala/io/citrine/lolo/validation/Merit.scala
bfolie: It might, since a real-world example would involve predicting the probability of simultaneously hitting N targets, where N could be higher than 2.

The generalization is straightforward, which is good. It might be a bit hairy to implement, but I think it's worth trying.

src/test/scala/io/citrine/lolo/bags/CorrelationStudy.scala
bfolie: I think that the quadratic function covers a similar case, in that it has no global correlation between the variables but strong local correlation.

That said, I'm open to suggestions if you think this or other test functions are meaningfully different.

src/main/scala/io/citrine/lolo/validation/Merit.scala
bfolie: I was thinking that the impact of a poor prediction levels off after a certain point. That is, being off by 10 standard deviations is really bad. Being off by 20 standard deviations is basically the same amount of really bad.

> Or is this accounted for by the PVA2D metric that uses a cutoff on the Mahalaonbis distance?

Are you referring to standard confidence? No. That measures how many predictions are within 1 (generalized) standard deviation. Being off by 10 standard deviations is the same as being off by 1.5.

bfolie:
Closing in favor of a stand-alone repo: https://github.com/CitrineInformatics/uncertainty-correlation-study/pull/1

PR #251, opened 2021-12-02
Multitask feature importance

bfolie:
PLA-8567

Implement `getFeatureImportance` for multitask model.

### DIFF
--- a/src/main/scala/io/citrine/lolo/trees/Nodes.scala
+++ b/src/main/scala/io/citrine/lolo/trees/Nodes.scala
@@ -199,7 +199,7 @@ class TrainingLeaf[T](
     * @return lightweight prediction node
     */
   def getNode(): ModelNode[PredictionResult[T]] = {
-    new ModelLeaf(leafLearner.train(trainingData).getModel().asInstanceOf[Model[PredictionResult[T]]], depth, trainingData.head._1.length)
+    new ModelLeaf(leafLearner.train(trainingData).getModel().asInstanceOf[Model[PredictionResult[T]]], depth, trainingData.size.toDouble)
   }
 
   override def getFeatureImportance(): mutable.ArraySeq[Double] = mutable.ArraySeq.fill(trainingData.head._1.size)(0.0)
--- a/src/main/scala/io/citrine/lolo/trees/Nodes.scala
+++ b/src/main/scala/io/citrine/lolo/trees/Nodes.scala
@@ -82,7 +82,6 @@ trait ModelNode[+T <: PredictionResult[Any]] extends Serializable {
   * @param split           to decide which branch to take
   * @param left            branch node
   * @param right           branch node
-  * @param numFeatures     number of features, used for Shapley computation
   * @param outputDimension dimension of model output, used for Shapley computation
   *                        1 for single-task regression, or equal to the number of classification categories.
   * @param trainingWeight  weight of training data in subtree (i.e. size of unweighted training set)
--- a/src/main/scala/io/citrine/lolo/trees/Nodes.scala
+++ b/src/main/scala/io/citrine/lolo/trees/Nodes.scala
@@ -92,7 +92,6 @@ class InternalModelNode[T <: PredictionResult[Any]](
                                                      split: Split,
                                                      left: ModelNode[T],
                                                      right: ModelNode[T],
-                                                     numFeatures: Int,
                                                      outputDimension: Int,
                                                      trainingWeight: Double
                                                    ) extends ModelNode[T] {
@@ -200,13 +199,13 @@ class TrainingLeaf[T](
     * @return lightweight prediction node
     */
   def getNode(): ModelNode[PredictionResult[T]] = {
-    new ModelLeaf(leafLearner.train(trainingData).getModel().asInstanceOf[Model[PredictionResult[T]]], depth, trainingData.head._1.length, trainingData.size.toDouble)
+    new ModelLeaf(leafLearner.train(trainingData).getModel().asInstanceOf[Model[PredictionResult[T]]], depth, trainingData.head._1.length)
   }
 
   override def getFeatureImportance(): mutable.ArraySeq[Double] = mutable.ArraySeq.fill(trainingData.head._1.size)(0.0)
 }
 
-class ModelLeaf[T](model: Model[PredictionResult[T]], depth: Int, numFeatures: Int, trainingWeight: Double) extends ModelNode[PredictionResult[T]] {
+class ModelLeaf[T](model: Model[PredictionResult[T]], depth: Int, trainingWeight: Double) extends ModelNode[PredictionResult[T]] {
   override def transform(input: Vector[AnyVal]): (PredictionResult[T], TreeMeta) = {
     (model.transform(Seq(input)), TreeMeta(depth))
   }
--- a/src/main/scala/io/citrine/lolo/trees/Nodes.scala
+++ b/src/main/scala/io/citrine/lolo/trees/Nodes.scala
@@ -34,7 +34,7 @@ abstract class TrainingNode[T <: AnyVal, S](
   def getFeatureImportance(): mutable.ArraySeq[Double]
 }
 
-trait ModelNode[T <: PredictionResult[Any]] extends Serializable {
+trait ModelNode[+T <: PredictionResult[Any]] extends Serializable {
   def transform(input: Vector[AnyVal]): (T, TreeMeta)
 
   /**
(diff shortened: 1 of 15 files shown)


src/main/scala/io/citrine/lolo/trees/Nodes.scala
```diff
@@ -92,7 +92,6 @@ class InternalModelNode[T <: PredictionResult[Any]](
                                                      split: Split,
                                                      left: ModelNode[T],
                                                      right: ModelNode[T],
-                                                     numFeatures: Int,
```
bfolie, line 95: `numFeature` parameter is unused here and in `ModelLeaf`

src/main/scala/io/citrine/lolo/trees/multitask/MultiTaskTrainingNode.scala
```diff
@@ -43,40 +76,36 @@ class MultiTaskTrainingNode(
...
-    } else if (rightChild.isDefined && right.nonEmpty) {
-      rightChild.get.getNode(index)
-    } else {
-      // If there are no children or the children don't have valid data for this label, emit a leaf (with GTM)
-      if (label.isInstanceOf[Double]) {
-        new TrainingLeaf[Double](reducedData.asInstanceOf[Seq[(Vector[AnyVal], Double, Double)]], new GuessTheMeanLearner(), 1).getNode()
-      } else {
-        new TrainingLeaf[Char](reducedData.asInstanceOf[Seq[(Vector[AnyVal], Char, Double)]], new GuessTheMeanLearner(), 1).getNode()
-      }
+    (leftChild, rightChild) match {
```
bfolie, line 79: this is just a refactoring from an if-else block to a match, which avoids the need for `.get` statements

src/main/scala/io/citrine/lolo/trees/multitask/MultiTaskTrainingNode.scala
```diff
@@ -29,9 +32,39 @@ class MultiTaskTrainingNode(
       (Some(new MultiTaskTrainingNode(leftData, randomizePivotLocation, rng = rng)), Some(new MultiTaskTrainingNode(rightData, randomizePivotLocation, rng = rng)))
   }
 
+  // get feature importance for the i'th label
+  def getFeatureImportance(index: Int): mutable.ArraySeq[Double] = {
```
bfolie, line 36: This is similar in structure to `getNode`. I tried to combine them into a single method that creates the model node and the feature importance simultaneously, but that led to strange errors in the sparse data tests. It would somehow wind up in a situation in which the only training data on a node were missing the desired label. I could not figure out the cause, and eventually gave up and reverted to two code blocks.

src/main/scala/io/citrine/lolo/trees/splits/Splitter.scala
```diff
@@ -17,4 +21,50 @@ object Splitter {
       Math.abs((x1 - x2) / (x1 + x2)) > 1.0e-9
     }
   }
+
```
bfolie, line 24: `RegressionSplitter`, `ClassificationSplitter`, and `MultitaskSplitter` all used essentially the same code for `getBestRealSplit`, so I centralized it here.

src/main/scala/io/citrine/lolo/trees/Nodes.scala
```diff
@@ -34,7 +34,7 @@ abstract class TrainingNode[T <: AnyVal, S](
   def getFeatureImportance(): mutable.ArraySeq[Double]
 }
 
-trait ModelNode[T <: PredictionResult[Any]] extends Serializable {
+trait ModelNode[+T <: PredictionResult[Any]] extends Serializable {
```
bfolie, line 37: Making the type covariant indicates that, for example, `InternalModelNode[PredictionResult[Double]]` is a subtype of `ModelNode[PredictionResult[Any]]`, and removes the need for a typecast.

src/main/scala/io/citrine/lolo/trees/Nodes.scala
sfriedowitz: The docstring says "used for Shapley computation". I guess that was a lie lol?

Also, if deleting this field, can you update the docstring please?

src/main/scala/io/citrine/lolo/trees/multitask/MultiTaskTrainingNode.scala
sfriedowitz: Since they are recursive calls I can see that being tricky. If you want to reduce code duplication a bit, this header block is identical:

```
    // Filter out "missing" values, which are NaN for regression and 0 for encoded categoricals
    val label = inputs.head._2(index)
    val reducedData = if (label.isInstanceOf[Double]) {
      inputs.map(x => (x._1, x._2(index).asInstanceOf[Double], x._3)).filterNot(_._2.isNaN)
    } else {
      inputs.map(x => (x._1, x._2(index).asInstanceOf[Char], x._3)).filter(_._2 > 0)
    }
    // Compute the valid data for each child
    val (left, right) = reducedData.partition(r => split.turnLeft(r._1))
```

It doesn't give you that much, but that could be branched into its own method for data filtering.

sfriedowitz: Also, my IDE is telling me that this type cast `reducedData.asInstanceOf[Seq[(Vector[AnyVal], Double, Double)]]` is redundant, and `reducedData` has the correct type already. Maybe that's another weird scala version thing?

src/main/scala/io/citrine/lolo/trees/splits/ClassificationSplitter.scala
```diff
@@ -37,7 +37,7 @@ case class ClassificationSplitter(randomizedPivotLocation: Boolean = false, rng:
 
       /* Use different spliters for each type */
       val (possibleSplit, possibleImpurity) = rep._1(index) match {
-        case _: Double => getBestRealSplit(data, calculator, index, minInstances)
+        case _: Double => Splitter.getBestRealSplit[Char](data, calculator, index, minInstances, randomizedPivotLocation, rng)
```
sfriedowitz, line 40: Is there any symmetry in `getBestCategoricalSplit` between the different subtypes that we can shuttle off to a static method? 

For instance, it seems that `getBestCategoricalSplit` on the `MultiTaskSplitter` is the most general, since it operates on an arbitrary number of distinct label types. Could that logic be adapted to provide a general method here?

src/main/scala/io/citrine/lolo/trees/Nodes.scala
bfolie: Gregor confirmed it was used in an early iteration but then taken out. I'll update the docstring.

src/main/scala/io/citrine/lolo/trees/multitask/MultiTaskTrainingNode.scala
bfolie: > It doesn't give you that much, but that could be branched into its own method for data filtering.

Every one of those values gets used later on, so you'd have to return all of them, which makes for a hideous method.

bfolie: > Also, my IDE is telling me that this type cast reducedData.asInstanceOf[Seq[(Vector[AnyVal], Double, Double)]] is redundant, and reducedData has the correct type already. Maybe that's another weird scala version thing?

My IDE tells me that too, but if I remove the typecast then it fails to compile.

src/main/scala/io/citrine/lolo/trees/splits/ClassificationSplitter.scala
bfolie: I looked into that a bit yesterday, but the codes seemed different in a more significant way, and it was the end of the day and my motivation for refactoring things had waned. 🤷

src/main/scala/io/citrine/lolo/trees/multitask/MultiTaskTrainingNode.scala
sfriedowitz: > Every one of those values gets used later on, so you'd have to return all of them, which makes for a hideous 
 method.

Yes you are right, pardon my quick take on that suggestion.

sfriedowitz (approved):
I wish for the sake of symmetry that `getBestCategoricalSplit` could also be refactored. But I don't see a clear way without changing the implementation between the methods quite a bit, so that should not block this.

PR #252, opened 2021-12-06
Finalize uncertainty correlation logic and interface

bfolie:
Now that we've tested the various methods of calculating correlated uncertainty, we can remove the unused methods and streamline the feature. This PR does the following:

* Adds an `observational: Boolean = true` argument to `getUncertaintyCorrelation`. This is to mirror `getUncertainty`
* Implement the bootstrap ensemble correlation when `observational = true` and the trivial method when `observational = false`
* Delete all code related to the Jackknife and Training Data uncertainty correlation methods
* Simplify the uncertainty correlation test


gregor-robinson (approved):
Big :+1: for deleting code

PR #253, opened 2021-12-08
Unify learner train interface

bfolie:
PLA-8798

`Learner` implements a method `train` that acts on a sequence of training data points, each of the form `(vector of inputs, label)`. But `MultiTaskLearner` implements `train` with a different interface: it takes one argument for the `inputs` and one argument for the `labels`. To make matters more confusing, the two arguments have different orientations. `inputs` has shape `(# training) x (# input dimensions)` while `labels` has shape `(# output dimensions) x (# training)`. This was undocumented and, at least to me, counter-intuitive.

In the context of Lolo it's not a big deal. But when I tried to implement the multi-task RF in Lolopy, I found that having these two different paradigms would lead to a lot of boilerplate, redundant code. So I decided to unify the interfaces, since I'm in here breaking things anyway.

`MultiTaskLearner` now implements train with a sequence of training data points, each of the form `(vector of inputs, vector of labels)`.

### DIFF
--- a/src/main/scala/io/citrine/lolo/bags/BaggedResult.scala
+++ b/src/main/scala/io/citrine/lolo/bags/BaggedResult.scala
@@ -403,9 +403,7 @@ case class MultiPredictionBaggedResult(
 case class MultiTaskBaggedResult(
                                   baggedPredictions: Seq[BaggedResult[Any]],
                                   realLabels: Seq[Boolean],
-                                  NibIn: Vector[Vector[Int]],
-                                  trainingLabels: Seq[Seq[Any]],
-                                  trainingWeights: Seq[Double]
+                                  NibIn: Vector[Vector[Int]]
                                 ) extends BaggedResult[Seq[Any]] with MultiTaskModelPredictionResult {
 
   /* transpose to be (# training) x (# models) */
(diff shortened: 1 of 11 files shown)


src/main/scala/io/citrine/lolo/bags/BaggedResult.scala
```diff
@@ -403,9 +403,7 @@ case class MultiPredictionBaggedResult(
 case class MultiTaskBaggedResult(
                                   baggedPredictions: Seq[BaggedResult[Any]],
                                   realLabels: Seq[Boolean],
-                                  NibIn: Vector[Vector[Int]],
-                                  trainingLabels: Seq[Seq[Any]],
-                                  trainingWeights: Seq[Double]
```
bfolie, line 408: These two arguments were being piped through to calculate the correlation coefficient using the training data. This is no longer necessary.

sfriedowitz (commented):
~~Should we add an overloaded `train` method for the known, single-task learners of the form `train(data: Seq[(Vector[Any], Any)])` ?~~

NVM. This already exists.

sfriedowitz (approved):

PR #255, opened 2021-12-10
Expose multitask random forest in Lolopy

bfolie:
PLA-8731

Makes it so that the `RandomForestRegressor` model in Lolopy can be trained on 2-dimensional label data, `y`. The model then returns predictions as a 2-dimensional array. This brings its behavior more in line with scikit-learn's random forest regressor, which can also operate on multiple labels. The following changes were made to enable this feature:
* The `RandomForest` learner in Lolo now checks to see if the output data is a `Seq` (previously it only checked whether or not it was a `Double`). If it's a `Seq`, then it produces a `MultiTaskBagger` around a standardized `MultiTaskTreeLearner`.
* The top-level `fit` method of `BaseLoloLearner` inspects the shape of the training labels `y`, and if it's a 2-d array then it gets sent to Java as an array instead of as a vector
* The `predict` method of `BaseLoloRegressor` changes the way it expects to receive predictions and uncertainties from Java, depending on the number of labels (for >1 labels, it reshapes the stream into a 2-dimensional array).
* The `predict` method now has a `return_cov_matrix` option in addition to the `return_std` option. Only one can be requested. `return_cov_matrix` returns an array of covariance matrices for each prediction.
* `MultiTaskTree`'s interface is expanded to include arguments for `minLeafInstances`, `maxDepth`, and `numFeatures`. 
* Although the random forest interface also includes a `leafLearner` argument, that one is not piped through to `MultiTaskTree`. If it is specified, then we print a message warning the user that it is being ignored. At this time, nothing except a guess-the-mean learner is sensible.

There are a few other small changes:
* I decided to upgrade scikit-learn to the recently-released version 1, on the belief that stable releases are better than unstable releases.
* This requires bumping the minimum python version from 3.6 to 3.7. Version 3.7 has been available since mid-2018.
* Scikit-learn v1 deprecates the boston housing dataset, so I replaced its use in tests with the diabetes dataset.
* The uncertainty for a multitask bagged model has been changed from `Option[Seq[Seq[Option[Any]]]]` to `Option[Seq[Seq[Any]]]`. The former was too clunky and it also made it more difficult to send the uncertainties over the wire. So what do we do if the uncertainty for a given label isn't defined? I decided to use `NaN`. This does have precedent in Lolo, which uses `NaN` and `null` to indicate missing labels in training data.

Despite many of the changes to Lolopy being at an abstract level, this PR does not enable any multitask models in Lolopy besides `RandomForestRegressor`.
* The linear regressor model in Lolo only accepts a single label
* The Extra Random forest model in Lolo makes use of customized extra-random regression/classification splitters, and those are not generalized to multitask models
* Lolopy makes a distinction between regression and classification learners, and so even though Lolo can train a mixed regression-classification multitask tree, that doesn't map onto Lolopy
* `RandomForestClassifier` would be the easiest to generalized to a multitask model. It would require modifying the `BaseLoloClassifier`'s `predict` and `proba` methods to handle the more complex data shape. This is a good follow-up candidate, especially since it would bring Lolpy more in line with sklearn, but it didn't seem high priority enough to delay the release.

### DIFF
--- a/python/lolopy/learners.py
+++ b/python/lolopy/learners.py
@@ -71,9 +71,10 @@ class BaseLoloLearner(BaseEstimator, metaclass=ABCMeta):
         learner = self._make_learner()
 
         # Determine the number of outputs
-        if len(y.shape) == 1:
+        y_shape = np.asarray(y).shape
+        if len(y_shape) == 1:
             self._num_outputs = 1
-        elif len(y.shape) == 2:
+        elif len(y_shape) == 2:
             self._num_outputs = y.shape[1]
         else:
             raise ValueError("Output array must be either 1- or 2-dimensional")
--- a/python/lolopy/learners.py
+++ b/python/lolopy/learners.py
@@ -3,7 +3,6 @@ from abc import abstractmethod, ABCMeta
 import numpy as np
 from lolopy.loloserver import get_java_gateway
 from lolopy.utils import send_feature_array, send_1D_array
-import random
 from sklearn.base import BaseEstimator, RegressorMixin, ClassifierMixin, is_regressor
 from sklearn.exceptions import NotFittedError
 
@@ -205,7 +204,6 @@ class BaseLoloRegressor(BaseLoloLearner, RegressorMixin):
         # If desired, return the uncertainty too
         if return_std:
             # TODO: This part fails on Windows because the NativeSystemBLAS is not found. Fix that
-            # TODO: This is only valid for regression models. Perhaps make a "LoloRegressor" class
             y_std_bytes = self.gateway.jvm.io.citrine.lolo.util.LoloPyDataLoader.getRegressionUncertainty(pred_result)
             y_std = np.frombuffer(y_std_bytes, 'float')
             return y_pred, y_std
--- a/python/lolopy/learners.py
+++ b/python/lolopy/learners.py
@@ -21,12 +21,12 @@ class BaseLoloLearner(BaseEstimator, metaclass=ABCMeta):
     In lolo, learners are not specific to a regression or classification problem and the type of problem is determined
     when fitting data is provided to the algorithm.
     In contrast, Scikit-learn learners for regression or classification problems are different classes.
-    We have implemented `BaseLoloRegressor` and `BaseLoloClassifier` abstract classes to make it easier for creating
+    We have implemented `BaseLoloRegressor` and `BaseLoloClassifier` abstract classes to make it easier to create
     a classification or regression version of a Lolo base class.
     The pattern for creating a scikit-learn compatible learner is to first implement the `_make_learner` and `__init__`
     operations in a special "Mixin" class that inherits from `BaseLoloLearner`, and then create a regression- or
     classification-specific class that inherits from both `BaseClassifier` or `BaseRegressor` and your new "Mixin".
-    See the RandomForest models as an example for this approach.
+    See the RandomForest models as an example of this approach.
     """
 
     def __init__(self):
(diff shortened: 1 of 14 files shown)


python/lolopy/tests/test_learners.py
```diff
@@ -134,7 +162,7 @@ def test_regression_tree(self):
         tree.max_depth = 2
         tree.fit(X, y)
         y_pred = tree.predict(X)
-        self.assertAlmostEqual(0.695574477, r2_score(y, y_pred))  # Result is deterministic
+        self.assertAlmostEqual(0.433370098, r2_score(y, y_pred))  # Result is deterministic
```
bfolie, line 165: different dataset, different R^2 value

python/lolopy/tests/test_learners.py
```diff
@@ -52,10 +52,10 @@ def test_rf_regressor(self):
         y_import = rf.get_importance_scores(X[:100, :])
         self.assertEqual((100, len(X)), y_import.shape)
 
-        # Basic test for functionality. R^2 above 0.98 was measured on 27Dec18
+        # Basic test for functionality. R^2 above 0.88 was measured on 2021-12-09
```
bfolie, line 55: different dataset, different R^2 value

src/main/scala/io/citrine/lolo/learners/ExtraRandomTrees.scala
```diff
@@ -52,23 +52,12 @@ case class ExtraRandomTrees(
     */
   override def train(trainingData: Seq[(Vector[Any], Any)], weights: Option[Seq[Double]]): TrainingResult = {
     val breezeRandBasis: RandBasis = new RandBasis(new MersenneTwister(rng.nextLong()))
+    val repOutput = trainingData.head._2
+    val isRegression = repOutput.isInstanceOf[Double]
+    val numFeatures: Int = RandomForest.getNumFeatures(subsetStrategy, trainingData.head._1.size, isRegression)
```
bfolie, line 57: The logic to compute the number of features was repeated across `RandomForest` and `ExtraRandomTrees`, so I consolidated it.

python/lolopy/learners.py
```diff
@@ -70,6 +70,14 @@ def fit(self, X, y, weights=None):
         # Instantiate the JVM object
         learner = self._make_learner()
 
+        # Determine the number of outputs
+        if len(y.shape) == 1:
```
latture, line 74: Can we be sure `y` is a NumPy array and has the `.shape` attribute (instead of something that's simply array-like)? Internally, sklearn uses [`check_array`](https://scikit-learn.org/stable/modules/generated/sklearn.utils.check_array.html), but maybe we could use something simpler like [`np.asarray`](https://numpy.org/doc/stable/reference/generated/numpy.asarray.html).

latture: Just saw the `np.array` call a few lines down.  Maybe we can move that ahead of the call to `.shape`?

bfolie: The `np.array` call occurs as part of `_convert_train_data`, which has to be called after this block because it relies on `self._num_outputs` being set. I'll use `np.asarray` before calling `.shape`.

latture:
> Although the random forest interface also includes a leafLearner argument, that one is not piped through to MultiTaskTree. If it is specified, then we print a message warning the user that it is being ignored. At this time, nothing except a guess-the-mean learner is sensible.

How hard would it be to check if it's guess the mean learner and throw an error otherwise?

bfolie:
> How hard would it be to check if it's either not defined or a guess-the-mean learner and throw an error otherwise?

It would be easy to do that instead of the `println` warning

latture (approved):
I might lean towards throwing an exception if the learner is defined incorrectly, but it's not a blocking concern.

latture (approved):

PR #258, opened 2021-12-17
Pass rng to splitters in RandomForest

bfolie:
When training a typical random forest model, there are two primary sources of randomness: bagging and selecting features to consider at each split. Previously we did not pass the `rng` to the splitter, so the latter was not reproducible. This PR fixes that.

### DIFF
--- a/src/main/scala/io/citrine/lolo/learners/RandomForest.scala
+++ b/src/main/scala/io/citrine/lolo/learners/RandomForest.scala
@@ -90,7 +90,7 @@ case class RandomForest(
           maxDepth = maxDepth,
           minLeafInstances = minLeafInstances,
           randomizePivotLocation = randomizePivotLocation,
-          rng = new Random(rng.nextLong())
+          rng = rng
         ))
         val bagger = MultiTaskBagger(
            if (randomlyRotateFeatures) MultiTaskFeatureRotator(DTLearner) else DTLearner,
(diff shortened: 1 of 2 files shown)


src/main/scala/io/citrine/lolo/learners/RandomForest.scala
```diff
@@ -68,8 +68,8 @@ case class RandomForest(
           numFeatures = numFeatures,
           minLeafInstances = minLeafInstances,
           maxDepth = maxDepth,
-          splitter = RegressionSplitter(randomizePivotLocation),
-          rng = rng
+          splitter = RegressionSplitter(randomizePivotLocation, rng = new Random(rng.nextLong())),
+          rng = new Random(rng.nextLong())
```
bfolie, line 72: It would be simpler to pass `rng` directly in both of these spots, but it seems like it is cleaner to split off two separate rngs for each piece. Am I overthinking this? @gregor-robinson

sfriedowitz: Can you just construct the `rng` outside the match arm and pass it to both?

sfriedowitz: Oh I am dumb, there already is an `rng` outside the match arm. I guess I don't see why we can't just pass that `rng` down?

bfolie: I'm probably over-thinking it and could just pass the `rng` down. But I'd like to check with Gregor

sfriedowitz (approved):

Issue #212, opened 2020-03-12
Use BaggerHelper in Bagger


gregor-robinson:
BaggerHelper, introduced in https://github.com/CitrineInformatics/lolo/pull/210, aims to subsume functionality common to Bagger and MultiTaskBagger so that the shared functionality can be maintained in one place. It is currently used only by MultiTaskBagger, so a subsequent PR should swing over functionality from Bagger as well.

PR #260, opened 2022-01-05
Bugfix/rescale ratios

bfolie:
When training a bagged model, if `uncertaintyCalibration` is true then a rescaling ratio is calculated. This ratio is used to rescale the bootstrap standard deviation into a reasonable estimate of the observational standard deviation. It is calculated by computing the out-of-bag error for each prediction, dividing by the bootstrap standard deviation, ordering, and taking the 68-percentile value. Assuming the out-of-bag errors are a reasonable estimate of the error on unknown points, then this means that, post-rescaling, approximately 68% of observations will lie within 1 error bar.

If `uncertaintyCalibration` is false then the ratio is 1.0, and the observational standard deviation is equal to the tree standard deviation, which is known to overestimate the uncertainty.

These rescale ratios were not being passed through when training a multitask random forest. This PR fixes that bug.

### DIFF
--- a/src/main/scala/io/citrine/lolo/bags/MultiTaskBagger.scala
+++ b/src/main/scala/io/citrine/lolo/bags/MultiTaskBagger.scala
@@ -113,7 +113,7 @@ class MultiTaskBaggedTrainingResult(
                                      rescaleRatios: Seq[Double]
                                    ) extends MultiTaskTrainingResult {
 
-  lazy val model = new MultiTaskBaggedModel(models, Nib, useJackknife, biasModels)
+  lazy val model = new MultiTaskBaggedModel(models, Nib, useJackknife, biasModels, rescaleRatios)
 
   // Each entry is a tuple, (feature vector, seq of predicted labels, seq of actual labels).
   // The labels are of type Option[Any] because a given training datum might not have a value for every single label.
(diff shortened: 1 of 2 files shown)


bfolie:
I think we have an inconsistency in our defaults.

When getting uncertainty, the [default](https://github.com/CitrineInformatics/lolo/blob/main/src/main/scala/io/citrine/lolo/PredictionResult.scala#L24) is to get the observational uncertainty, which for a bagged model means calculating the bootstrap standard deviation and re-scaling.

But the default for [RandomForest](https://github.com/CitrineInformatics/lolo/blob/main/src/main/scala/io/citrine/lolo/learners/RandomForest.scala#L41) is to set `uncertaintyCalibration = false`, which means that the rescaling ratio isn't calculated and users are getting the uncorrected bootstrap standard deviation. This is also the default for [Bagger](https://github.com/CitrineInformatics/lolo/blob/main/src/main/scala/io/citrine/lolo/bags/Bagger.scala#L31) and [MultiTaskBagger](https://github.com/CitrineInformatics/lolo/blob/main/src/main/scala/io/citrine/lolo/bags/MultiTaskBagger.scala#L26).

I think that these defaults should be trained so that `uncertaintyCalibration = true`.

sfriedowitz (approved):
This looks good for the sake of this PR and the bugfix.

This is beyond this PR, but I was checking out the code a bit for `Bagger` and it seems like there is a large amount of code duplication between that and `BaggerHelper` which is used for the multi task model. I haven't tried to compare whether a straight refactor would be possible, or if there are some other tricky points, but it seems like using `BaggerHelper` within `Bagger` in a future refactor would cleanup alot of duplication.

bfolie:
> I was checking out the code a bit for `Bagger` and it seems like there is a large amount of code duplication between that and `BaggerHelper` which is used for the multi task model. I haven't tried to compare whether a straight refactor would be possible, or if there are some other tricky points, but it seems like using `BaggerHelper` within `Bagger` in a future refactor would cleanup alot of duplication.

There's actually an issue for this: #212 . I just haven't gotten around to it

PR #263, opened 2022-01-06
Uncertainty calibration default true

### LINKED ISSUE #212: Use BaggerHelper in Bagger

bfolie:
Sets `uncertaintyCalibration` to have a default value of true for `RandomForest`, `Bagger`, and `MultiTaskBagger`*. This is because the default uncertainty for the resulting predictions is observational uncertainty, which makes the most sense as a recalibrated bootstrap standard deviation, but that recalibration is not done if `uncertaintyCalibration = false`. Therefore, our defaults led to nonintuitive behavior.

The rescale ratio can sometimes be Infinity if the number of training data and bags are small. Since we have two tests that run thousands of trials and check that sigma is positive, now that `uncertaintyCalibration` is true it sometimes runs into that situation and throws an error. I have made is so that `ratio` is set to 1.0 in this scenario.

Since I was changing the rescale ratio logic, I took the opportunity to consolidate the `BaggerHelper` code, which was duplicated in `Bagger. This resolves #212.

*The default is still false for `ExtraRandomTrees`, since this learner disables the bootstrap by default.

### DIFF
--- a/src/main/scala/io/citrine/lolo/bags/Bagger.scala
+++ b/src/main/scala/io/citrine/lolo/bags/Bagger.scala
@@ -28,7 +28,7 @@ case class Bagger(
                    numBags: Int = -1,
                    useJackknife: Boolean = true,
                    biasLearner: Option[Learner] = None,
-                   uncertaintyCalibration: Boolean = false,
+                   uncertaintyCalibration: Boolean = true,
                    disableBootstrap: Boolean = false,
                    randBasis: RandBasis = Rand
                  ) extends Learner {
--- a/src/main/scala/io/citrine/lolo/bags/Bagger.scala
+++ b/src/main/scala/io/citrine/lolo/bags/Bagger.scala
@@ -114,67 +114,23 @@ case class Bagger(
       .reduce(Bagger.combineImportance)
       .map(_.map(_ / importances.size))
 
-    /* Out-of-bag error and uncertainty for each point, calculating by combining the result of each tree.
-    Define as lazy so we only compute them if they are needed for the ratio or bias learner calculation */
-    lazy val oobErrors: Seq[(Vector[Any], Double, Double)] = trainingData.indices.flatMap { idx =>
-      val oobModels = models.zip(Nib.map(_ (idx))).filter(_._2 == 0).map(_._1).asInstanceOf[ParSeq[Model[PredictionResult[Double]]]]
-      if (oobModels.size < 2) {
-        None
-      } else {
-        Async.canStop()
-        val model = new BaggedModel(oobModels, Nib.filter {
-          _ (idx) == 0
-        }, useJackknife, disableBootstrap = disableBootstrap)
-        val predicted = model.transform(Seq(trainingData(idx)._1))
-        val error = predicted.getExpected().head - trainingData(idx)._2.asInstanceOf[Double]
-        val uncertainty = predicted match {
-          case x: RegressionResult => x.getStdDevObs.get.head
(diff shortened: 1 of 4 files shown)


src/main/scala/io/citrine/lolo/bags/Bagger.scala
```diff
@@ -114,64 +114,23 @@ case class Bagger(
...
-        case (_, 0.0, 0.0) => 1.0
-        case (_, _, 0.0) => Double.PositiveInfinity
-        case (_, error, uncertainty) => Math.abs(error / uncertainty)
-      }.sorted.drop((oobErrors.size * 0.68).toInt).head
-    } else {
-      1.0
-    }
-    assert(!ratio.isNaN && !ratio.isInfinity, s"Uncertainty calibration ratio is not real: $ratio")
 
     /* Wrap the models in a BaggedModel object */
```
sfriedowitz, line 118: I think we can avoid the nested conditionals and make this a bit cleaner as follows:

```
    /* Wrap the models in a BaggedModel object */
    val helper = BaggerHelper(models, trainingData, Nib, useJackknife, uncertaintyCalibration)
    val biasModel = if (biasLearner.isDefined && helper.oobErrors.nonEmpty && helper.isRegression) {
      Async.canStop()
      Some(biasLearner.get.train(helper.biasTraining).getModel().asInstanceOf[Model[PredictionResult[Double]]])
    } else {
      None
    }

    if (helper.isRegression) {
      new BaggedTrainingResult[Double](models.asInstanceOf[ParSeq[Model[PredictionResult[Double]]]], averageImportance, Nib, trainingData, useJackknife, biasModel, helper.ratio, disableBootstrap)
    } else {
      new BaggedTrainingResult[Any](models, averageImportance, Nib, trainingData, useJackknife, None, helper.ratio, disableBootstrap)
    }
```

That is how I tackled it when going to make this change. Also, we can remove the `isRegression` defined above in the method and just use `helper.isRegression`.

sfriedowitz (commented):
One comment below!

src/main/scala/io/citrine/lolo/bags/Bagger.scala
bfolie: yeah that's better

sfriedowitz (approved):
LGTM

src/main/scala/io/citrine/lolo/bags/Bagger.scala
```diff
@@ -114,65 +109,35 @@ case class Bagger(
...
-      } else {
-        new BaggedTrainingResult[Any](models, averageImportance, Nib, trainingData, useJackknife, None, ratio, disableBootstrap)
-      }
+      new BaggedTrainingResult[Any](
+        models = models,
+        featureImportance = averageImportance,
+        Nib = Nib,
+        trainingData = trainingData,
+        useJackknife = useJackknife,
+        biasModel = biasModel,
```
sfriedowitz, line 137: Oh I just realized this. `biasModel` will always be None in this branch. Should we just explicit set it `None` ?

bfolie: I thought about that, but decided it would be bug prone because it smears the "what's the bias model" logic out into two places.

PR #267, opened 2022-01-07
Enable scalafmt code formatting

sfriedowitz:
Re-doing https://github.com/CitrineInformatics/lolo/pull/265 where I caused some issues with docstring formatting.


.travis.yml
```diff
@@ -24,6 +24,7 @@ jobs:
   include:
   - stage: test
     script:
+    - sbt +scalafmtCheckAll
```
latture, line 27: Do we need the `+` here?

sfriedowitz: I'm not sure. We have it for `+test`, but the `travis` documentation was very ambiguous about it. Do you know for sure?

sfriedowitz: The few examples here don't seem to have a `+argument` used, so I'm not sure why we have `+test` to begin with?

https://docs.travis-ci.com/user/languages/scala/

latture: We need it for `test` because we want to run the tests across all cross-compiled scala versions. It's probably not necessary here, but it doesn't hurt either.

latture (approved):

bfolie (approved):

PR #275, opened 2022-02-05
Move to 2.0

sfriedowitz:


bfolie (approved):
All looks reasonable, assuming the build passes...

sfriedowitz:
Hmm I have no clue why the build is failing. It's something related to the `netlib` compile. I'll have to dig in more, possibly next week.

sfriedowitz:
I also have a PR open to upgrade `theta` to 2.0 here: https://github.com/CitrineInformatics/theta/pull/20

I wonder if its dependency hell related to that.

project/Dependencies.scala
```diff
@@ -4,19 +4,19 @@ object Dependencies {
...
 
   val loloDeps = Seq(
-    "com.github.fommil.netlib" % "all" % netlibVersion pomOnly(),
-    "junit" % "junit" % junitVersion % "test",
-    "org.scalanlp" %% "breeze" % breezeVersion,
-    "io.citrine" %% "theta" % thetaVersion,
-    "org.scalatest" %% "scalatest" % scalaTestVersion % "test",
-    "com.novocode" % "junit-interface" % "0.11" % "test",
-    "org.knowm.xchart" % "xchart" % "3.5.2"
+    "com.github.fommil.netlib" % "all"             % netlibVersion pomOnly (),
```
bfolie, line 14: could this extra space after `pomOnly` be causing a problem?

sfriedowitz: `scalafmt` is enforcing it being there, so I don't think thats the issue

bfolie:
> I also have a PR open to upgrade `theta` to 2.0 here: [CitrineInformatics/theta#20](https://github.com/CitrineInformatics/theta/pull/20)
> 
> I wonder if its dependency hell related to that.

The error message does sound like conflicting netlibs, so I am reasonably optimistic that it will work once the newer version of theta is pulled in.

sfriedowitz:
The issue was related to the fact that we run `sbt ci-release`, which compiles a fat jar with all dependencies via the command `sbt assembly`. This SBT command requires a merge strategy when compiling all dependencies into the fat jar, to handle duplicate files coming from different dependencies. Up to now, we used the default merge strategy without customizing it at all, which fails on duplicates.

The new breeze version brings in a new netlib implementation, [as mentioned here.](https://github.com/scalanlp/breeze#:~:text=various%20native%20libraries.-,As%20of%20Breeze%201.3,-%2C%20we%20now%20use). This was what was causing duplicate file issues. To fix it, I edited the build settings to take the first instance of duplicate `.class` files, which is what is [generally recommended here](https://github.com/sbt/sbt-assembly#:~:text=more%20settings%20here%20...%0A%20%20)-,Merge%20Strategy,-If%20multiple%20files).

project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=1.3.8
```
sfriedowitz, line 1: We were just on a very old version, so figured I would update it while mucking around the build settings.

project/Dependencies.scala
sfriedowitz: @latture Do you know if we need this dependency anymore?

Breeze >= 1.4 moved to a new version of netlib, not the `fommil` version. I am unsure what purpose this serves. But I do know that you dealt with this issue a few months ago.

sfriedowitz:
Just to confirm, I built the assembly jar and tested `lolopy` with it using the example notebooks, and it all appears to work fine.

project/Dependencies.scala
sfriedowitz: I have not. I should try that before we decide on merging this into Lolo.

Lolo was previously using breeze 1.2, so maybe this upgrade still will not conflict with `ecos` ?

latture: Hopefully there won't be any issues, but it's worth checking just to be safe.

bfolie (approved):

sfriedowitz:
@bfolie I have to wait on this until we cut a new `theta` release with the old `fommil` netlib removed. But I'm just bogged down by other things today. I'll re-request when ready.

bfolie (approved):

PR #276, opened 2022-02-15
Does installing BLAS stop the travis job from stalling?

sfriedowitz:
We shall see


sfriedowitz:
Everything I wrote above there was wrong. However, [this page](https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received) instructed me how to fix the time outs.

PR #278, opened 2022-02-15
SeedRandom for test suite

sfriedowitz:
The other issue I've noticed build failures were just random flakes. I whipped up a seed random scheme for the test suite this morning to try to fix that. 

The one thing I recognize is that this `SeedRandomMixin` is not currently using a splittable RNG, although I don't know if we care about that for the test suite.

### DIFF
--- a/src/test/scala/io/citrine/lolo/bags/MultiTaskBaggerTest.scala
+++ b/src/test/scala/io/citrine/lolo/bags/MultiTaskBaggerTest.scala
@@ -54,12 +54,11 @@ class MultiTaskBaggerTest extends SeedRandomMixIn {
     */
   @Test
   def testBaggedMultiTaskGetUncertainty(): Unit = {
-    rng.setSeed(37895L) // This seed is required for test to pass consistently *shrugs*
     val noiseLevel = 100.0
 
     Seq(MultiTaskTreeLearner()).foreach { baseLearner =>
       // These are in Seqs as a convenience for repurposing this test as a diagnostic tool.
-      Seq(64).foreach { nRows =>
+      Seq(128).foreach { nRows =>
         Seq(16).foreach { nCols =>
           Seq(2).map { n => n * nRows }.foreach { nBags =>
             // Used for error output.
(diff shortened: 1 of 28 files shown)


src/test/scala/io/citrine/lolo/bags/MultiTaskBaggerTest.scala
```diff
@@ -57,6 +54,7 @@ class MultiTaskBaggerTest {
     */
   @Test
   def testBaggedMultiTaskGetUncertainty(): Unit = {
+    rng.setSeed(37895L) // This seed is required for test to pass consistently *shrugs*
```
sfriedowitz, line 57: This test only passes consistently with this seed, which is what was originally here. I didn't spend the time trying to figure out what was going on.

bfolie: This test is training a multi-task RF on 20 different training sets, each consisting of nothing but noise in the output. It applies each model to 4 inputs and calculates the uncertainty with `observational = true` and `observational = false`. It then checks, effectively, that the uncertainty is always higher with `observational = true`. The requirement is framed in terms of a Jeffrey's prior and a cutoff level, but with these parameters the effective result is that the test fails if the observational sigma is ever lower than the mean sigma. Sometimes this happens once, just barely.

We can fix it by increasing the number of rows from 64 to 128, which increases the number of bags. Basically, the jackknife calculation is noisy, and the more bags we add the more it calms down. Notice that the [comparable test for single-task models](https://github.com/CitrineInformatics/lolo/blob/0018527327184868c40d08bef7aec35015ee3976/src/test/scala/io/citrine/lolo/bags/BaggedResultTest.scala#L49) uses 128 rows and it passes every time.

sfriedowitz: tis' fixed

bfolie (approved):
This is a significant improvement. Using a splittable rng would make the tests even more repeatable, but that's not something we need to worry about until we implement a splittable rng as part of the main code.

PR #282, opened 2022-05-11
Tolerant equality to zero in rescale ratio calculation

sfriedowitz:
This is in response to the following bug: https://citrine.atlassian.net/browse/PLA-9529

The training data had many values equal to zero. This led to roundoff errors in the OOB error/uncertainty that made those values both very close to zero, but not quite zero. The ultimately led to the rescale ratio calculating `error / uncertainty` with uncertainty very close to zero, leading to large ratios, and subsequently, large variance of the predicted values. This manifested in poor CV results with our live models.

This PR adds in a tolerant comparison for the OOB errors to zero when calculating rescale ratio.

### DIFF
--- a/src/main/scala/io/citrine/lolo/bags/Bagger.scala
+++ b/src/main/scala/io/citrine/lolo/bags/Bagger.scala
@@ -135,7 +135,7 @@ case class Bagger(
         trainingData = trainingData,
         useJackknife = useJackknife,
         biasModel = biasModel,
-        rescale = helper.ratio,
+        rescale = helper.rescaleRatio,
         disableBootstrap = disableBootstrap
       )
     } else {
@@ -146,7 +146,7 @@ case class Bagger(
         trainingData = trainingData,
         useJackknife = useJackknife,
         biasModel = biasModel,
-        rescale = helper.ratio,
+        rescale = helper.rescaleRatio,
         disableBootstrap = disableBootstrap
       )
     }
--- a/src/main/scala/io/citrine/lolo/bags/Bagger.scala
+++ b/src/main/scala/io/citrine/lolo/bags/Bagger.scala
@@ -4,7 +4,7 @@ import breeze.linalg.DenseMatrix
 import breeze.stats.distributions.{Poisson, Rand, RandBasis}
 import io.citrine.lolo.stats.metrics.{ClassificationMetrics, RegressionMetrics}
 import io.citrine.lolo.util.{Async, InterruptibleExecutionContext}
-import io.citrine.lolo.{Learner, Model, PredictionResult, RegressionResult, TrainingResult}
+import io.citrine.lolo.{Learner, Model, PredictionResult, TrainingResult}
 
 import scala.collection.parallel.ExecutionContextTaskSupport
 import scala.collection.parallel.immutable.{ParRange, ParSeq}
--- a/src/main/scala/io/citrine/lolo/bags/BaggerHelper.scala
+++ b/src/main/scala/io/citrine/lolo/bags/BaggerHelper.scala
@@ -1,6 +1,6 @@
 package io.citrine.lolo.bags
 
-import io.citrine.lolo.stats.MathUtils
+import io.citrine.lolo.stats.{MathUtils, StatsUtils}
 import io.citrine.lolo.util.Async
 import io.citrine.lolo.{Model, PredictionResult, RegressionResult}
 
@@ -61,7 +61,9 @@ protected case class BaggerHelper(
   /** Uncertainty calibration ratio based on the OOB errors. */
   val rescaleRatio: Double = if (uncertaintyCalibration && isRegression && useJackknife) {
     Async.canStop()
-    BaggerUtil.calculateRescaleRatio(oobErrors.map { case (_, e, u) => (e, u) })
+    val trainingLabels = trainingData.collect { case (_, x: Double) if !(x.isInfinite || x.isNaN) => x }
+    val zeroTolerance = StatsUtils.range(trainingLabels) / 1e12
+    BaggerHelper.calculateRescaleRatio(oobErrors.map { case (_, e, u) => (e, u) }, zeroTolerance = zeroTolerance)
   } else {
     1.0
   }
@@ -79,7 +81,7 @@ protected case class BaggerHelper(
   }
 }
 
-object BaggerUtil {
+object BaggerHelper {
 
   /**
     * Calculate the uncertainty calibration ratio, which is the 68th percentile of error/uncertainty for the training points.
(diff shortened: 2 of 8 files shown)


src/main/scala/io/citrine/lolo/bags/Bagger.scala
```diff
@@ -146,7 +146,7 @@ case class Bagger(
         trainingData = trainingData,
         useJackknife = useJackknife,
         biasModel = biasModel,
-        rescale = helper.ratio,
+        rescale = helper.rescaleRatio,
```
sfriedowitz, line 149: I felt like this name is a bit better since we call this parameter `rescale` in the other parts of the code. It only involved changes in the `Bagger` file, so was not large to rename it.

src/main/scala/io/citrine/lolo/bags/BaggerUtil.scala
```diff
@@ -57,38 +58,62 @@ protected case class BaggerHelper(
...
+    * meaning there may only be 2 out-of-bag models.
+    *
+    * @param oobErrors sequence of tuples of (error, uncertainty) from OOB predictions for each training row
+    * @return uncertainty rescale ratio
+    */
+  def calculateRescaleRatio(oobErrors: Seq[(Double, Double)]): Double = {
+    val oneSigmaRatio = oobErrors
+      .map {
+        case (error, uncertainty) =>
+          val errorIsZero = MathUtils.tolerantEquals(error, 0.0)
```
sfriedowitz, line 99: Without tolerant equality, we were checking if `error == 0.0` and `uncertainty == 0.0` to see if we return a value of `1.0` for the ratio. This check was missing when `error ~ 1e-16` or `uncertainty ~ 1e-16`, leading to large ratios when dividing the two values.

src/test/scala/io/citrine/lolo/bags/BaggerUtilTest.scala
```diff
@@ -0,0 +1,17 @@
+package io.citrine.lolo.bags
+
+import io.citrine.lolo.SeedRandomMixIn
+import org.junit.Test
+
+class BaggerUtilTest extends SeedRandomMixIn {
+
+  /** Test that the rescale ratio is tolerant to error/uncertainty close to zero. */
+  @Test
+  def testRescaleRatioPrecision(): Unit = {
```
sfriedowitz, line 10: This test would fail before these changes.

src/main/scala/io/citrine/lolo/bags/BaggerUtil.scala
scott-isaacson: Is there a risk here if the values of the training data are really small (e.g. on the order of 1e-12)? Or is the scale of the training data not relevant because we wrap the the `lolo` learners in a `Standardizer`? I suppose the worst case would be that the rescale ratio would be set to `1`.

sfriedowitz: In practical use cases we care about, I don't think the scale is relevant because we are wrapping each learner in a `Standardizer`. For a general solution for `lolo` as a standalone software package, this might be relevant. 

We could perhaps try to determine the tolerance based on the scale of the training data, but that was something I was going to hammer out via this PR review.

src/main/scala/io/citrine/lolo/bags/BaggerHelper.scala
```diff
@@ -1,5 +1,6 @@
 package io.citrine.lolo.bags
```
sfriedowitz, line 1: I renamed the file to align with what the objects are actually called.

src/main/scala/io/citrine/lolo/bags/BaggerHelper.scala
```diff
@@ -57,38 +58,63 @@ protected case class BaggerHelper(
...
+    */
+  def calculateRescaleRatio(oobErrors: Seq[(Double, Double)], zeroTolerance: Double = 1e-12): Double = {
+    val oneSigmaRatio = oobErrors
+      .map {
+        case (error, uncertainty) =>
+          val errorIsZero = MathUtils.tolerantEquals(error, 0.0, zeroTolerance)
+          val uncertaintyIsZero = MathUtils.tolerantEquals(uncertainty, 0.0, zeroTolerance)
+          if (errorIsZero && uncertaintyIsZero) {
+            1.0
+          } else {
```
sfriedowitz, line 106: There was previously a check here that if `uncertainty = 0.0`, then we return `Double.PositiveInfinity`. But that is automatically handled by the 2nd branch of this `if-statement`, because we already determined that `error != 0.0`, so `anything / 0.0 = Double.PositiveInfinity`, and scala is fine with that math (it won't return `NaN`).

bfolie:
The core change in this PR is that if the error and uncertainty are both infinitesimal rounding errors, then the rescale ratio for that bag is counted as 1. Without this change then you might get some large or small value like 1000 or 1e-3. If this occurs for most of the bags, then Mithril will wind up returning uncertainties that are way too large or way too small.

I don't see any problems with this change, but it's not clear to me that it's the underlying cause of PLA-9529, or indeed would occur in production.

First, I'm not sure how we'd get values like 1e-12 for both the error and uncertainty. The predicted value is the average of the individual tree predictions, each of which are a training data point. Even if every tree predicts the same value, averaging 64 copies of 0.0 should yield 0.0 exactly. I don't think that this is a situation in which roundoff errors arise. Similarly for the uncertainty.

Second, I'm not sure how a bad rescale ratio leads to the negative predicted value seen in the ticket. I suppose it could be an artifact of the large uncertainty + sampling. If the prediction is 0.1 +/- 1000, then the mean sample value could be less than 0.

How did you establish that zero-adjacent values were occurring in this case for `error` and `uncertainty`?

sfriedowitz:
> I don't see any problems with this change, but it's not clear to me that it's the underlying cause of PLA-9529, or indeed would occur in production.

This is verified occurring in production, and required extensive debugging to get to the root cause of it.

> How did you establish that zero-adjacent values were occurring in this case for error and uncertainty?

It was a deep dive through the workbench on their live problem. Ultimately, we came to realize that substantially large uncertainty estimates were leading to the poor evaluation results. I was able to track down through the debugger that the rescale ratio was contributing to the huge uncertainty. 

Then, coming over into Lolo, was able to verify near-zero values of error/uncertainty were leading to the large rescale ratio.  It is related to wrapping the base learner in a `Standardizer`, which apparently imbues some amount of floating point error during the rescaling.

> I suppose it could be an artifact of the large uncertainty + sampling. If the prediction is 0.1 +/- 1000, then the mean sample value could be less than 0.

I believe this is exactly what was going on.

bfolie:
> It is related to wrapping the base learner in a Standardizer, which apparently imbues some amount of floating point error during the rescaling.

Ah, yes. I see how the Standardizer could lead to roundoff errors. Good use of the Workbench.

bfolie (approved):

PR #284, opened 2022-05-31
PLA-9689: Evaluate linear regression and linear ensemble

sfriedowitz:
I started out here by looking into whether zero weights in the linear model were causing instabilities in the normal equations. As it turns out, that is **not** the case. Regardless of whether you filter out zero weight rows or not, the problem is unchanged.

Although, for high-dimensional linear regression the LHS of the normal equations is often low rank and non-invertible. Adding regularization **always** solves that (see slide 22 [here](https://people.orie.cornell.edu/mru8/orie4741/lectures/underdetermined_lsq.pdf)). To produce stable linear models within Mithril, we simply are going to have to add regularization in all cases.

So, what does this PR do then? A couple of things:
1) General refactor of the linear regression learner for style/readability
2) Uses the `\` operator for inversions, which is more stable and performant. Internally, this will either perform a cholesky or an SVD depending on the problem structure, so it can be used for both the fully determined and underdetermined cases.
3) Wraps the underdetermined case in the `Try` block, because solution can throw if the internal SVD does not converge.


src/main/scala/io/citrine/lolo/linear/LinearRegression.scala
```diff
@@ -26,80 +28,77 @@ case class LinearRegressionLearner(
...
+    } getOrElse {
+      val totalWeight = weights.map(_.sum).getOrElse(numSamples.toDouble)
+      val mean = sum(yw) / totalWeight // weighted mean of training labels
+      if (fitIntercept) {
+        (DenseVector.zeros[Double](numFeatures - 1), mean)
+      } else {
+        (DenseVector.zeros[Double](numFeatures), mean)
       }
-    } else {
-      pinv(A) * b
```
sfriedowitz, line 87: This was unwrapped before, which often threw exceptions if the internal SVD did not converge.

src/main/scala/io/citrine/lolo/linear/LinearRegression.scala
```diff
@@ -26,80 +28,77 @@ case class LinearRegressionLearner(
...
+    val yw = weightsMatrix.map(W => W * y).getOrElse(y)
+
+    val (coefficients, intercept) = Try {
+      // If regularized/overdetermined, attempt to solve the full normal equations
+      // If underdetermined, fall back to a least-squares solution that uses SVD/pseudo-inverse internally
+      val beta = if (regularized || numSamples >= numFeatures) {
+        val l = math.pow(lambda, 2) * DenseVector.ones[Double](numFeatures)
+        if (fitIntercept) l(-1) = 0.0
+        val A = Xt * Xw + diag(l)
+        val b = Xt * yw
```
gregor-robinson, line 74: Coming back to your question on zoom: yes, it is correct to use the weighted observations.

gregor-robinson (changes requested):

src/main/scala/io/citrine/lolo/linear/LinearRegression.scala
```diff
@@ -26,80 +28,77 @@ case class LinearRegressionLearner(
...
+
+    val (coefficients, intercept) = Try {
+      // If regularized or overdetermined, design matrix is (possibly) invertible
+      // If underdetermined, fall back to a least-squares solution that uses SVD/pseudo-inverse internally
+      val beta = if (regularized || numSamples >= numFeatures) {
+        val l = math.pow(lambda, 2) * DenseVector.ones[Double](numFeatures)
+        if (fitIntercept) l(-1) = 0.0
+        val A = Xt * Xw + diag(l)
+        val b = Xt * yw
+        A \ b
```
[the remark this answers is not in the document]
sfriedowitz: It looks like breeze actually calls an `LU` internally for square matrices: https://github.com/scalanlp/breeze/blob/4c0610366764a20b19a98fc1f7c6c5c16cd308e6/math/src/main/codegen/breeze/linalg/operators/DenseMatrixOps.scala

Nonetheless, I think the `\` call is best practice, and it is faster than a direct `pinv` call.

src/main/scala/io/citrine/lolo/linear/LinearRegression.scala
```diff
@@ -26,82 +29,86 @@ case class LinearRegressionLearner(
...
-      .filter(_._1.isInstanceOf[Double])
-      .filterNot(_._1.asInstanceOf[Double].isNaN)
-      .map(_._2)
-      .filterNot(i => trainingData.exists(_._1(i).asInstanceOf[Double].isNaN))
-      .filterNot { i =>
-        val unregularized = !regParam.exists(_.asInstanceOf[Double] > 0.0)
-        lazy val constant = trainingData.forall(_._1(i) == trainingData.head._1(i))
-        unregularized && constant // remove constant features if there's no regularization
+    val indices = rep.zipWithIndex
+      .collect { case (x: Double, idx) if !x.isNaN => idx }
```
bfolie, line 40: Is this redundant? It's removing features for which the representative row has a value of NaN. But then the next block removes features for which any row has a value of NaN.

sfriedowitz: At first I was leaving it in because I thought it would save us a list traversal by pre-filtering. But, I actually think it causes an extra traversal over the `rep` because `hasNans.exists` will exit early on satisfaction, and `constant` is lazy. 

So yeah, its redundant and should be removed.

sfriedowitz: Ehh, the filter/collect logic turns out a bit cleaner by pre-collecting the `Double` indices, so I'm going to leave as is.

src/main/scala/io/citrine/lolo/linear/LinearRegression.scala
```diff
@@ -76,19 +79,22 @@ case class LinearRegressionLearner(
...
-      }
+    } match {
+      case Success(beta) =>
+        if (fitIntercept) {
+          (beta(0 to -2), beta(-1))
+        } else {
+          (beta, 0.0)
+        }
+      case Failure(e) =>
+        logger.error(s"Encountered an exception solving normal equations: ${e.getLocalizedMessage}")
```
gregor-robinson, line 90: Error log level is high. Is this situation significant enough to warrant automatic engineering attention, or should we log at a lower level and wait for customer feedback when it really should be working?

sfriedowitz: I can move it to `warn`.

sfriedowitz: That aligns with how we log internally in other places.

gregor-robinson (approved):

PR #286, opened 2022-06-02
Use time -l to get maximum memory usage against current config

teekennedy:


teekennedy:
Looks like the tests on Travis are CPU bound. In order to make a real difference in running them faster, we'll need to pay to upgrade the size of our Travis CI VMs.

Closing this.

PR #288, opened 2022-09-29
Reproducible training

bfolie:
The intention of this PR is to allow complete reproducibility in Lolo. Given the same training data and the same configuration, we should be able to produce identical models when starting from the same random seed. The biggest change made to accomplish this goal is to **move random seeds from objects to methods**. Currently, a random seed can be specified when the user creates a `RandomForest` object. This is the same pattern as scikit-learn, but I have changed the random seed to be an argument of the `RandomForest`'s `train` method. There are several reasons.

1. Locality makes debugging easier. Putting the random seed on the function makes each function call reproducible, independent of global state. This allows us to drill down into the behavior of a specific piece of code when trying to understand an error.
2. It's more flexible. The same persisted configuration can be retrained with a specific random seed if we need to reproducible an earlier result. But it can also be retrained with an arbitrary seed.
3. It facilitates determinism in a multi-threaded environment. Logic can be fanned out to parallel processes and as long as each function call gets the same random seed, the result is deterministic. If the seed is on the objects then we need to construct a new object holding that seed in order to pass to each parallel process.

The clearest user-facing change is that learners, like `RandomForest` no longer have a random seed and the `train` method does have a random seed. But there are several internal methods that now have a random seed:
* getting the best split
* hyperparameter optimization
* running cross-validation and other model quality tests
* randomly rotating features

I'm submitting this PR as-is because it accomplishes the goal and is quite large already. But there are a few things to change in follow-on PRs.
* Currently, a multi-task RF is not able to pass the random seed all the way down to training the individual leaf learners. It's not an issue in practice because the leaf learners are deterministic, but it's still a potential bug. We'll need to re-architect the multi-task node creation structure to make this possible.
* Similarly, the regression and classification node structure leads to random seeds in some object constructors. These are internal objects that are created as part of training and expose no methods to the outside, so it's not an immediate issue. But it would be better to change how they are structured.

I also removed Scala 2.12 and upgraded the Travis build file from using jdk8 to jdk11

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,8 +1,8 @@
 import Dependencies._
 
 ThisBuild / name := "lolo"
-ThisBuild / scalaVersion := "2.13.4"
-ThisBuild / crossScalaVersions := List("2.13.4", "2.12.4")
+ThisBuild / scalaVersion := "2.13.8"
+ThisBuild / crossScalaVersions := List("2.13.8")
 ThisBuild / organization := "io.citrine"
 ThisBuild / organizationName := "Citrine Informatics"
 ThisBuild / homepage := Some(url("https://github.com/CitrineInformatics/lolo"))
@@ -25,15 +25,6 @@ ThisBuild / sonatypeRepository := "https://s01.oss.sonatype.org/service/local"
 ThisBuild / pomIncludeRepository := { _ => false }
 
 ThisBuild / libraryDependencies ++= logging ++ loloDeps
-ThisBuild / libraryDependencies ++= {
-  CrossVersion.partialVersion(scalaVersion.value) match {
-    case Some((2, major)) if major <= 12 =>
-      Seq()
-    case _ =>
-      Seq("org.scala-lang.modules" %% "scala-parallel-collections" % scalaParallelCollectionsVersion)
-  }
-}
-
 ThisBuild / Test / testOptions += Tests.Argument(TestFrameworks.JUnit, "-v")
 
 // Assembly settings
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -2,6 +2,7 @@ import sbt._
 
 object Dependencies {
   lazy val thetaVersion = "1.1.5"
+  lazy val sprandomVersion = "0.1.0"
   lazy val breezeVersion = "2.0"
   lazy val junitVersion = "4.13.1"
   lazy val scalaTestVersion = "3.2.2"
@@ -17,11 +18,13 @@ object Dependencies {
   )
 
   lazy val loloDeps = Seq(
-    "io.citrine"      %% "theta"           % thetaVersion,
-    "org.scalanlp"    %% "breeze"          % breezeVersion,
-    "junit"            % "junit"           % junitVersion     % "test",
-    "org.scalatest"   %% "scalatest"       % scalaTestVersion % "test",
-    "com.github.sbt"   % "junit-interface" % "0.13.3"         % "test",
-    "org.knowm.xchart" % "xchart"          % "3.5.2"
+      "io.citrine"            %% "theta"                      % thetaVersion,
+    "io.citrine"              %% "sprandom"                   % sprandomVersion,
+    "org.scalanlp"            %% "breeze"                     % breezeVersion,
+    "org.scala-lang.modules"  %% "scala-parallel-collections" % scalaParallelCollectionsVersion,
+    "junit"                    % "junit"                      % junitVersion     % "test",
+    "org.scalatest"           %% "scalatest"                  % scalaTestVersion % "test",
+    "com.github.sbt"           % "junit-interface"            % "0.13.3"         % "test",
+    "org.knowm.xchart"         % "xchart"                     % "3.5.2"
   )
 }
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -2,12 +2,10 @@ import sbt._
 
 object Dependencies {
   lazy val thetaVersion = "1.1.5"
-  lazy val sprandomVersion = "0.1.0"
+  lazy val sprandomVersion = "0.1.1"
   lazy val breezeVersion = "2.0"
   lazy val junitVersion = "4.13.1"
   lazy val scalaTestVersion = "3.2.2"
-
-  // this is used in build.sbt so that it works with both 2.12 and 2.13
   lazy val scalaParallelCollectionsVersion = "1.0.0"
 
   lazy val logging = Seq(
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -16,7 +16,7 @@ object Dependencies {
   )
 
   lazy val loloDeps = Seq(
-      "io.citrine"            %% "theta"                      % thetaVersion,
+    "io.citrine"              %% "theta"                      % thetaVersion,
     "io.citrine"              %% "sprandom"                   % sprandomVersion,
     "org.scalanlp"            %% "breeze"                     % breezeVersion,
     "org.scala-lang.modules"  %% "scala-parallel-collections" % scalaParallelCollectionsVersion,
--- a/python/lolopy/tests/test_learners.py
+++ b/python/lolopy/tests/test_learners.py
@@ -32,7 +32,7 @@ def _make_linear_data():
 class TestRF(TestCase):
 
     def test_rf_regressor(self):
-        rf = RandomForestRegressor(random_seed = 31247895)
+        rf = RandomForestRegressor()
 
         # Train the model
         X, y = load_diabetes(return_X_y=True)
@@ -42,7 +42,7 @@ class TestRF(TestCase):
             rf.predict(X)
 
         # Fit the model
-        rf.fit(X, y)
+        rf.fit(X, y, random_seed=31247895)
 
         # Run some predictions
         y_pred = rf.predict(X)
@@ -79,13 +79,13 @@ class TestRF(TestCase):
         self.assertIsNone(rf.model_)
 
     def test_rf_multioutput_regressor(self):
-        rf = RandomForestRegressor(random_seed=810355)
+        rf = RandomForestRegressor()
         # A regression dataset with 3 outputs
         X, y = load_linnerud(return_X_y=True)
         num_data = len(X)
         num_outputs = y.shape[1]
 
-        rf.fit(X, y)
+        rf.fit(X, y, random_seed=810355)
         y_pred, y_std = rf.predict(X, return_std=True)
         _, y_cov = rf.predict(X, return_cov_matrix=True)
 
@@ -103,11 +103,11 @@ class TestRF(TestCase):
             rf.predict(X, return_std=True, return_cov_matrix=True)
 
     def test_classifier(self):
-        rf = RandomForestClassifier(random_seed = 34789)
+        rf = RandomForestClassifier()
 
         # Load in the iris dataset
         X, y = load_iris(return_X_y=True)
-        rf.fit(X, y)
+        rf.fit(X, y, random_seed=34789)
 
         # Predict the probability of membership in each class
         y_prob = rf.predict_proba(X)
@@ -126,7 +126,7 @@ class TestRF(TestCase):
         self.assertAlmostEqual(acc, 1)  # Given default settings, we should get perfect fitness to training data
 
     def test_serialization(self):
-        rf = RandomForestClassifier(random_seed = 234785)
+        rf = RandomForestClassifier()
 
         # Make sure this doesn't error without a model training
         data = pkl.dumps(rf)
@@ -134,7 +134,7 @@ class TestRF(TestCase):
 
         # Load in the iris dataset and train model
         X, y = load_iris(return_X_y=True)
-        rf.fit(X, y)
+        rf.fit(X, y, random_seed=234785)
 
         # Try saving and loading the model
         data = pkl.dumps(rf)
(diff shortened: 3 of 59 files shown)


build.sbt
```diff
@@ -25,15 +25,6 @@ ThisBuild / sonatypeRepository := "https://s01.oss.sonatype.org/service/local"
...
 ThisBuild / libraryDependencies ++= logging ++ loloDeps
-ThisBuild / libraryDependencies ++= {
-  CrossVersion.partialVersion(scalaVersion.value) match {
-    case Some((2, major)) if major <= 12 =>
-      Seq()
-    case _ =>
-      Seq("org.scala-lang.modules" %% "scala-parallel-collections" % scalaParallelCollectionsVersion)
-  }
-}
-
```
bfolie, line 36: This can go away if we're no longer supporting Scala 2.12

python/lolopy/tests/test_learners.py
```diff
@@ -32,7 +32,7 @@ def _make_linear_data():
 class TestRF(TestCase):
```
bfolie, line 32: I should add a reproducibility test from the LoloPy side

src/main/scala/io/citrine/lolo/trees/multitask/MultiTaskTrainingNode.scala
```diff
@@ -147,17 +151,20 @@ class MultiTaskTrainingNode(
       case (_, Some(theRightChild)) if right.nonEmpty =>
         theRightChild.getNode(index)
       case (_, _) =>
+        // TODO (PLA-10388): get the rng in here somehow (right now it's instantiating a random rng)
```
bfolie, line 154: This isn't an issue in practice because the typical base learner, `GuessTheMeanLearner`, trains deterministically. It needs to be fixed before the next release, but I'd like to do it in a follow-up PR.

src/test/scala/io/citrine/lolo/bags/BaggerTest.scala
```diff
@@ -242,80 +237,30 @@ class BaggerTest extends SeedRandomMixIn {
...
     val calibratedTime = 1.0e-9 * (System.nanoTime() - startAgain)
 
     assert(calibratedTime < unCalibratedTime, s"The calibration scheme has experienced a dramatic slowdown")
   }
 
-  /**
-    * Test that the bagged learner can be interrupted
-    */
-  @Test
-  def testInterrupt(): Unit = {
```
bfolie, line 275: Training is no longer interruptible

src/test/scala/io/citrine/lolo/learners/ExtraRandomTreesTest.scala
```diff
@@ -184,21 +185,40 @@ class ExtraRandomTreesTest extends SeedRandomMixIn {
...
-
-object ExtraRandomTreesTest {
-  def main(argv: Array[String]): Unit = {
-    new ExtraRandomTreesTest()
-      .testClassification()
-    new ExtraRandomTreesTest()
-      .testClassificationUnbiased()
+  /** Test that the same random seed leads to identical models */
+  @Test
+  def testReproducibility(): Unit = {
```
bfolie, line 198: Testing deterministic training for an Extra Random Forest

src/test/scala/io/citrine/lolo/learners/RandomForestTest.scala
```diff
@@ -102,6 +103,36 @@ class RandomForestTest extends SeedRandomMixIn {
     }
   }
 
+  /** Test that the same random seed leads to identical models */
+  @Test
```
bfolie, line 107: Testing deterministic training for a Random Forest

project/Dependencies.scala
```diff
@@ -17,11 +16,13 @@ object Dependencies {
   )
 
   lazy val loloDeps = Seq(
-    "io.citrine"      %% "theta"           % thetaVersion,
-    "org.scalanlp"    %% "breeze"          % breezeVersion,
-    "junit"            % "junit"           % junitVersion     % "test",
-    "org.scalatest"   %% "scalatest"       % scalaTestVersion % "test",
-    "com.github.sbt"   % "junit-interface" % "0.13.3"         % "test",
-    "org.knowm.xchart" % "xchart"          % "3.5.2"
+      "io.citrine"            %% "theta"                      % thetaVersion,
```
sfriedowitz, line 19: Formatting?

src/main/scala/io/citrine/lolo/Learner.scala
```diff
@@ -1,27 +1,32 @@
 package io.citrine.lolo
 
-/**
-  * Created by maxhutch on 11/14/16.
-  */
+import io.citrine.random.Random
```
sfriedowitz, line 3: This is not an issue with this package. But should the classpath from SpRandom be `io.citrine.sprandom` instead of `io.citrine.random` ?

src/main/scala/io/citrine/lolo/Learner.scala
```diff
@@ -1,27 +1,32 @@
...
     * @param trainingData to train on
     * @param weights      for the training rows, if applicable
+    * @param rng          random number generator for reproducibility
     * @return training result containing a model
     */
-  def train(trainingData: Seq[(Vector[Any], Any)], weights: Option[Seq[Double]] = None): TrainingResult
+  def train(
+      trainingData: Seq[(Vector[Any], Any)],
+      weights: Option[Seq[Double]] = None,
+      rng: Random = Random()
```
sfriedowitz, line 18: Do we want to use the `Random.default` method?

sfriedowitz: Edit: I forget why we even have `Random.default` vs `Random()`, both of which instantiate an effectively unseeded `Random`. I guess one was done before the other.

src/main/scala/io/citrine/lolo/bags/Bagger.scala
```diff
@@ -76,6 +76,7 @@ case class Bagger(
     )
 
     /* Compute the number of instances of each training row in each training sample */
+    val randBasis = new RandBasis(new ThreadLocalRandomGenerator(new MersenneTwister(rng.nextLong())))
```
sfriedowitz, line 79: Add a helper method in this package for getting a breeze `RandBasis` from a `Random` ?

src/main/scala/io/citrine/lolo/bags/Bagger.scala
```diff
@@ -101,19 +102,19 @@ case class Bagger(
...
-      val sampleWeights = Nib(i).zip(weightsActual).map(p => p._1.toDouble * p._2)
-
-      // Train the model
-      val meta = method.train(trainingData.toVector, Some(sampleWeights))
-
-      // Extract the model and feature importance from the TrainingResult
-      (meta.getModel(), meta.getFeatureImportance())
-    }.unzip
+    // Learn the actual models in parallel
+    val iterator = Nib.indices.toVector
```
sfriedowitz, line 106: This variable is not an iterator so the naming seems off. But also, I don't think we need to predefine it here. Could we just do `rng.zip(Nib.indices)` ?

src/main/scala/io/citrine/lolo/bags/Bagger.scala
```diff
@@ -124,7 +125,9 @@ case class Bagger(
     val helper = BaggerHelper(models, trainingData, Nib, useJackknife, uncertaintyCalibration)
     val biasModel = if (biasLearner.isDefined && helper.oobErrors.nonEmpty && helper.isRegression) {
       Async.canStop()
```
sfriedowitz, line 127: Is this necessary anymore? Is the `Async` object necessary at all anymore?

src/main/scala/io/citrine/lolo/bags/MultiTaskBagger.scala
```diff
@@ -53,6 +54,7 @@ case class MultiTaskBagger(
     val actualBags = if (numBags > 0) numBags else trainingData.size
 
     // Compute the number of instances of each training row in each training sample
+    val randBasis = new RandBasis(new ThreadLocalRandomGenerator(new MersenneTwister(rng.nextLong())))
```
sfriedowitz, line 57: Helper method would be nice, as mentioned above.

src/main/scala/io/citrine/lolo/bags/MultiTaskBagger.scala
```diff
@@ -67,12 +69,18 @@ case class MultiTaskBagger(
       .toVector
     val weightsActual = weights.getOrElse(Seq.fill(trainingData.size)(1.0))
 
-    val parIterator = new ParRange(Nib.indices)
-    parIterator.tasksupport = new ExecutionContextTaskSupport(InterruptibleExecutionContext())
-    val (models: ParSeq[MultiTaskModel], importances: ParSeq[Option[Vector[Double]]]) = parIterator.map { i =>
-      val meta = method.train(trainingData, Some(Nib(i).zip(weightsActual).map(p => p._1.toDouble * p._2)))
-      (meta.getModel(), meta.getFeatureImportance())
-    }.unzip
+    val iterator = Nib.indices.toVector
```
sfriedowitz, line 72: Same comment on the iterator and predefinition.

src/main/scala/io/citrine/lolo/trees/multitask/MultiTaskTrainingNode.scala
sfriedowitz: What interface needs to be adjusted to allow this?

sfriedowitz (commented):
Great changes to the package. I may have missed some, but I recommending adding a util method for mapping the `Random` to a breeze RandBasis. 

Also, do we have an explicit test here checking serializability? Would be good to have if not.

src/main/scala/io/citrine/lolo/Learner.scala
bfolie: .....yeah, it probably should

src/main/scala/io/citrine/lolo/Learner.scala
bfolie: Different idioms for doing the same thing 🤷

src/main/scala/io/citrine/lolo/bags/Bagger.scala
bfolie: Will do

src/main/scala/io/citrine/lolo/bags/Bagger.scala
bfolie: We can't do `rng.zip(Nib.indices)` because `Nib.indices` is a `Range` and doesn't have an implicit `buildFrom` method. Which seemed strange to me because I think that said method has an obvious implementation. But it still seemed like a bunch of boilerplate and it would be easier to construct the iterable directly.

The reason I put it on a separate line is because I think it reads more easily to say "here's the thing we're iterating over and now we're zipping it with `rng`" rather than "we're zipping an `rng` with _this blob you now have to think about_." But I agree that the name `iterator` is incorrect.

src/main/scala/io/citrine/lolo/bags/Bagger.scala
bfolie: No. I'm going to pull out `ASync` in a follow-up PR. There are several things I want to do before we release v5 😄

bfolie:
> Also, do we have an explicit test here checking serializability? Would be good to have if not.

The serializability of a Random Forest is checked as part of [LoloPyDataLoaderTest](https://github.com/CitrineInformatics/lolo/blob/main/src/test/scala/io/citrine/lolo/util/LoloPyDataLoaderTest.scala#L121)

src/main/scala/io/citrine/lolo/stats/StatsUtils.scala
```diff
@@ -58,4 +62,7 @@ object StatsUtils {
     if (sigma2X == 0 || sigma2Y == 0) return 0.0
     covariance(X, Y, Some(actualWeights)) / math.sqrt(sigma2X * sigma2Y)
   }
+
+  def breezeRandBasis(rng: Random): RandBasis =
```
sfriedowitz, line 66: ya know what I just realized. We have roughly this same function within `TestUtils.getBreezeRandBasis`. 

if its easy we should delete that one, and do a find/replace to use this one.

sfriedowitz (approved):
LGTM

src/main/scala/io/citrine/lolo/trees/multitask/MultiTaskTrainingNode.scala
bfolie: At a high level I think that the training node classes need to be rethought of as being like Learners. Even if they don't inherit explicitly from the `Learner` trait, we should think about them as having a configuration to which we pass data and then when queried they perform logic to calculate the split and produce the model node. Right now the split calculation happens upon instantiation and the model node creation happens later when `getNode` is called. That separation is, I think, fundamentally why we can't thread the `rng` through properly.

Because the "configuration" and "training" steps here are inherently intertwined I think in practice it makes more sense to combine them. Instead of having `getNode` be a function that performs logic, the `node` will be a value in the `TrainingNode` case class. This means `TrainingNode` can't be instantiated directly. Instead it will have a builder method and _this method will take the rng_. It will use that rng to calculate the split, which has an element of randomness. It will then call the builder methods of the left and right nodes (if they exist) and pass the rng to those methods. Then it will return the node, which is now a case class that knows everything already and does no calculation.

This is half-baked but it's what I'll try first.

tl;dr: all of the `Node`s will have their constructors change and they will get builder methods.

BAMcvoelker:
Great! Thanks so much. I am looking forward to trying this out!

PR #289, opened 2022-10-05
Deterministic training nodes

bfolie:
This PR brings the tree leaves and nodes in line with the pattern established in #288: instead of including `rng`s in case classes they are passed as part of builder methods. The stochastic computation is confined to the builder method and the resulting training node is deterministic. This makes the code easier to reason about and also fixes a potential determinism bug for multitask forests.

The logic for the regression and classification nodes is straight-forward and nearly identical. The logic for a multitask node is more complicated because a global tree is formed based on all training data, but the path through the tree is label-dependent. At a given node, one label may stop and evaluate a leaf model while another label may continue along a split. I handle this problem with a sequence of "label-wise instructions" that also serve to carry the leaf models.

Other changes:
* Make some classes into case classes
* Change the `getTrainingWeight` method so that it sums weights instead of counting unique rows
* Remove an unused type parameter from `TrainingNode`
* Pass a splitter to `MultiTaskTree` instead of the `randomizePivotLocation` flag


src/main/scala/io/citrine/lolo/trees/Nodes.scala
```diff
@@ -2,30 +2,26 @@ package io.citrine.lolo.trees
...
-  * @param remainingDepth to stop growing the node
-  * @tparam T type of the input vector
-  * @tparam S type of the model output
+  * @tparam T type of the model output
   */
-abstract class TrainingNode[T <: AnyVal, S](
-    trainingData: Seq[(Vector[T], S, Double)],
-    remainingDepth: Int = Int.MaxValue
-) extends Serializable {
+trait TrainingNode[+T] extends Serializable {
```
bfolie, line 14: It was unnecessary to have a type parameter for the input and another type parameter for the label, since in practice the inputs are never constrained past `AnyVal`. So now there is only one type parameter, corresponding to the label.

src/main/scala/io/citrine/lolo/trees/Nodes.scala
```diff
@@ -181,38 +177,20 @@ class InternalModelNode[T <: PredictionResult[Any]](
...
-    trainingData: Seq[(Vector[AnyVal], T, Double)],
-    leafLearner: Learner,
-    depth: Int,
-    rng: Random
-) extends TrainingNode(
-      trainingData = trainingData,
-      remainingDepth = 0
-    ) {
+/** A leaf defined by a training result. */
+trait TrainingLeaf[T] extends TrainingNode[T] {
```
bfolie, line 181: Previously there was this concrete `TrainingLeaf` class that was being used by the classification tree, and also a `RegressionTrainingLeaf` class that was very similar but specialized for regression.
I think that this high-level `Node` file should be limited to traits/abstract classes or implementations that are truly universal. So I redefined this as a trait and implemented a `ClassificationTrainingLeaf`. In practice, the two training leaves are identical except for their feature importance calculation.

src/main/scala/io/citrine/lolo/trees/classification/ClassificationTrainingNode.scala
```diff
@@ -1,152 +1,99 @@
...
-  * @param numFeatures      number of features used in selecting splits
-  * @param minLeafInstances minimum training instances per node
-  * @param remainingDepth   to stop growing the node
-  * @param maxDepth         maximum depth of tree
-  * @param numClasses       total number of classes, used as an output dimensionality
-  * @param splitter         used to select splits
-  * @param rng              random number generator for reproducibility
-  */
-class ClassificationTrainingNode(
+case class ClassificationTrainingNode(
```
bfolie, line 8: This is one important change (and the corresponding change to `RegressionTrainingNode`). The case class is now defined with its child nodes and the splitting logic is done in a builder that takes an `rng`.

src/main/scala/io/citrine/lolo/trees/multitask/MultiTaskTrainingNode.scala
```diff
@@ -2,171 +2,207 @@ package io.citrine.lolo.trees.multitask
...
-class MultiTaskTrainingNode(
-    inputs: Seq[(Vector[AnyVal], Array[AnyVal], Double)],
-    numFeatures: Int,
-    maxDepth: Int,
-    minInstances: Int,
-    randomizePivotLocation: Boolean = false,
-    rng: Random
+case class MultiTaskTrainingNode(
+    trainingData: Seq[(Vector[AnyVal], Array[AnyVal], Double)],
+    labelWiseInstructions: Seq[MultiTaskLabelInstruction],
```
bfolie, line 15: `MultiTaskTrainingNode` presents additional complications because the tree traversal logic is label-dependent. A split might exist, but if there are no more training rows for some label at this node, then we must terminate and make a prediction.
The solution here is to include a sequence of instructions saying what to do for each label -- take the split, evaluate a given leaf model, go left, or go right. This is how we can smuggle the label-wise models in to various points on the tree. But there's a flaw in this approach that I have not been able to get around -- there's nothing fundamental preventing an inconsistency between the instructions and the child nodes. For example, what if the label instruction is `DoSplit()` but the left/right node options are `None`?
In practice, the logic of the builder method prevents this from happening. But it's still a flaw.

src/main/scala/io/citrine/lolo/trees/multitask/MultiTaskTree.scala
```diff
@@ -13,13 +14,13 @@ import io.citrine.lolo.{Model, MultiTaskLearner, MultiTaskTrainingResult, Parall
...
   * @param minLeafInstances minimum number of training instances per leaf
-  * @param randomizePivotLocation whether to generate splits randomly between the data points
+  * @param splitter to determine the best split given data
   */
 case class MultiTaskTreeLearner(
     numFeatures: Int = -1,
     maxDepth: Int = 30,
     minLeafInstances: Int = 1,
-    randomizePivotLocation: Boolean = false
+    splitter: MultiTaskSplitter = MultiTaskSplitter(randomizePivotLocation = true)
```
bfolie, line 23: It's more generic to specify the splitter

src/main/scala/io/citrine/lolo/trees/regression/RegressionTrainingLeaf.scala
```diff
@@ -1,37 +1,25 @@
...
 import io.citrine.random.Random
 
 import scala.collection.mutable
 
-/**
-  * Training leaf node for regression trees
-  */
-class RegressionTrainingLeaf(
+/** Training leaf node for regression trees. */
+case class RegressionTrainingLeaf(
```
bfolie, line 10: Made this a case class instead of a class

src/main/scala/io/citrine/lolo/trees/splits/Split.scala
```diff
@@ -5,7 +5,7 @@ import scala.collection.BitSet
 /**
   * Splits are used by decision trees to partition the input space
   */
-abstract trait Split extends Serializable {
+sealed trait Split extends Serializable {
```
bfolie, line 8: Made these case classes

src/main/scala/io/citrine/lolo/trees/classification/ClassificationTrainingNode.scala
```diff
@@ -1,152 +1,99 @@
...
-          splitter,
-          rng
+  ): TrainingNode[Char] = {
+    val sufficientData = trainingData.size >= 2 * minLeafInstances &&
+      remainingDepth > 0 &&
+      trainingData.exists(_._2 != trainingData.head._2)
+    if (sufficientData) {
+      val (split: Split, deltaImpurity: Double) =
+        splitter.getBestSplit(trainingData, numFeatures, minLeafInstances, rng)
+      if (!split.isInstanceOf[NoSplit]) {
```
sfriedowitz, line 59: A match statement seems more appropriate here

src/main/scala/io/citrine/lolo/trees/classification/ClassificationTree.scala
```diff
@@ -73,38 +73,29 @@ case class ClassificationTreeLearner(
...
-        maxDepth = maxDepth,
-        minLeafInstances = minLeafInstances,
-        numClasses = trainingData.map { _._2 }.distinct.length,
-        splitter = splitter,
-        rng = rng
-      )
-    }
-
-    /* Wrap them up in a regression tree */
+    // Recursively build the tree via it snodes and wrap the top node in a ClassificationTreeTrainingResult
```
sfriedowitz, line 76: Minor typo `its nodes`

src/main/scala/io/citrine/lolo/trees/multitask/MultiTaskTrainingNode.scala
```diff
@@ -2,171 +2,207 @@ package io.citrine.lolo.trees.multitask
...
-  */
-class MultiTaskTrainingNode(
-    inputs: Seq[(Vector[AnyVal], Array[AnyVal], Double)],
-    numFeatures: Int,
-    maxDepth: Int,
-    minInstances: Int,
-    randomizePivotLocation: Boolean = false,
-    rng: Random
+case class MultiTaskTrainingNode(
+    trainingData: Seq[(Vector[AnyVal], Array[AnyVal], Double)],
```
sfriedowitz, line 14: Why are we using an `Array[AnyVal]` for the labels type? Seems odd to be passing around an immutable `Vector` on one end and the mutable collection on the other if we're not using that aspect of the collection.

src/main/scala/io/citrine/lolo/trees/multitask/MultiTaskTrainingNode.scala
```diff
@@ -2,171 +2,207 @@ package io.citrine.lolo.trees.multitask
...
   }
 }
+
+/**
+  * An enumeration specifying what behavior to perform for a specific label.
+  * The logic of MultiTaskTrainingNode.build() enforces that instructions will only be given in valid situations.
+  * For example, DoSplit() will only occur if there is a valid split and this label has training data that
+  * go down both paths.
+  */
+sealed trait MultiTaskLabelInstruction
```
sfriedowitz, line 193: I have personally developed the taste of organizing ADTs/enums within the companion object of the class they're associated with. So this could be added within the above object for `MultiTaskTrainingNode`. But it's a personal preference thing so not a blocker.

src/main/scala/io/citrine/lolo/trees/splits/Split.scala
```diff
@@ -41,7 +41,7 @@ class NoSplit extends Split {
     *
     * @return index of the input vector used by this split
     */
-  override def getIndex(): Int = -1
+  override def index: Int = -1
```
sfriedowitz, line 44: Would it be more appropriate to have the index be an `Option[Int]` to account for the fact that a split may not have a valid index?

Probably a ton of work to update that now b/c of its widespread usages in the code, but seems like a more appropriate interface.

sfriedowitz (commented):
I guess the main question is what do for the multi task instructions problem. I left a comment there that may help, but I may need to get some more info from you to leave a more insightful suggestion.

src/main/scala/io/citrine/lolo/trees/Nodes.scala
```diff
@@ -2,30 +2,26 @@ package io.citrine.lolo.trees
...
+  * @tparam T type of the model output
   */
-abstract class TrainingNode[T <: AnyVal, S](
-    trainingData: Seq[(Vector[T], S, Double)],
-    remainingDepth: Int = Int.MaxValue
-) extends Serializable {
+trait TrainingNode[+T] extends Serializable {
+
+  // TODO (PLA-10433): make this a structured type
+  def trainingData: Seq[(Vector[AnyVal], T, Double)]
```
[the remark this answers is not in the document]
bfolie: We never constrain the input type, so it wasn't providing any value

src/main/scala/io/citrine/lolo/trees/multitask/MultiTaskTrainingNode.scala
bfolie: I don't know; that's historical. I'll change it

src/main/scala/io/citrine/lolo/trees/splits/Split.scala
bfolie: Using -1 to mean "no split" isn't great. But I don't think an option would be better because then you allow a situation in which an actual split has no index, and `index` should _always_ be defined for split objects.

I think a better structure would be to have a `trait SplitResult` which is extended by `case class NoSplit extends SplitResult` and `trait Split` extends `SplitResult`. The trait `Split` would have an `index` and it would be extended by `RealSplit` and `CategoricalSplit`. So you could match on the `SplitResult` to establish whether the split was successful, and then you know that `index` is well-defined.

But that's a bit more than I want to take on right now.

src/main/scala/io/citrine/lolo/trees/multitask/MultiTaskTrainingNode.scala
bfolie: That could work. It means that the split info will be duplicated several times. I'll try it out.

src/main/scala/io/citrine/lolo/trees/multitask/MultiTaskTrainingNode.scala
bfolie: I know you do, but I like them as separate traits/case classes

src/main/scala/io/citrine/lolo/trees/multitask/MultiTaskTrainingNode.scala
```diff
@@ -2,171 +2,207 @@ package io.citrine.lolo.trees.multitask
...
+        trainingData,
+        numFeatures,
+        minInstances,
+        rng
+      )
+    } else {
+      (NoSplit(), 0.0)
+    }
+
+    val exampleRow = trainingData.head
```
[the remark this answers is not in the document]
bfolie: I don't think it works well as its own method because you have to pass so much context (about 6-7 arguments) that it ends up being pretty bulky. I added comments to segment the logic.

src/main/scala/io/citrine/lolo/trees/multitask/MultiTaskTrainingNode.scala
bfolie: Changed it. There's some duplication but I think it's better overall.

src/main/scala/io/citrine/lolo/trees/classification/ClassificationTrainingLeaf.scala
```diff
@@ -0,0 +1,35 @@
...
+import scala.collection.mutable
+
+/** Training leaf for a classification tree. */
+case class ClassificationTrainingLeaf(
+    trainingData: Seq[(Vector[AnyVal], Char, Double)],
+    trainingResult: TrainingResult,
+    depth: Int
+) extends TrainingLeaf[Char] {
+
+  override def getFeatureImportance(): mutable.ArraySeq[Double] = mutable.ArraySeq.fill(trainingData.head._1.size)(0.0)
```
sfriedowitz, line 16: Should we rename this to just `featureImportance` to follow the other style changes?

sfriedowitz: Also, why is this and associated methods on multi-task returning a `mutable.ArraySeq[Double]`. Was it always like that? If there's a strong reason for this collection choose we should keep it, but if not, i think we should use a standard immutable vector as we do elsewhere.

src/main/scala/io/citrine/lolo/trees/multitask/MultiTaskTrainingNode.scala
```diff
@@ -2,171 +2,195 @@ package io.citrine.lolo.trees.multitask
...
-    val (left, right) = reducedData.partition(r => split.turnLeft(r._1))
-
-    // get feature importance from the children if they exist, or from this node if it is a leaf
-    (leftChild, rightChild) match {
-      case (Some(theLeftChild), Some(theRightChild)) if left.nonEmpty && right.nonEmpty =>
-        val ans =
-          theLeftChild.getFeatureImportance(index).zip(theRightChild.getFeatureImportance(index)).map(p => p._1 + p._2)
-        ans(split.getIndex()) = ans(split.getIndex()) + deltaImpurity
+  // get feature importance for the `index`th label
+  def featureImportanceByLabelIndex(index: Int): mutable.ArraySeq[Double] = {
```
sfriedowitz, line 20: Same question on the choice of collection returned by this method.

src/main/scala/io/citrine/lolo/trees/classification/ClassificationTrainingLeaf.scala
bfolie: It was always like this because the feature importance is built up by starting with an array of 0s and repeatedly adding contributions, then normalizing at the end.

bfolie: > Should we rename this to just featureImportance to follow the other style changes?

yes

sfriedowitz (approved):
LGTM

PR #290, opened 2022-10-05
remove Async

bfolie:
It's an artifact from when Lolo was running in a synchronous web service


sfriedowitz (approved):
All deletions, love it. LGTM pending tests pass

PR #291, opened 2022-10-06
Update dependencies and patterns

bfolie:
In preparation for releasing Lolo v5.0.0, update to Scala 2.13 and Python 3.8. Also update most dependencies and plugins to their most recent versions.

Removing support for Scala 2.12 allows us to remove some deprecated patterns. Specifically, `zipped`, `toIterable`, `mapValues`, and Java converters.


sfriedowitz (approved):
LGTM

PR #294, opened 2022-10-11
PLA-10433: Generic typing on Learner and TrainingResult interfaces

sfriedowitz:


src/test/scala/io/citrine/lolo/DataGenerator.scala
```diff
@@ -0,0 +1,113 @@
+package io.citrine.lolo
+
+import breeze.stats.variance
+import io.citrine.lolo.linear.LinearRegressionLearner
+import io.citrine.lolo.stats.functions.Friedman
+import io.citrine.random.Random
+
+object DataGenerator {
```
sfriedowitz, line 8: This stuff is pretty much all copied over from `TestUtils`.

The entire purpose of the `TrainingData` interface is to allow us to generate these rows of training data and perform the binning operation on them while maintaining strong typing on the labels. The previous `TestUtils.binTrainingData` function did not allow for this and was incompatible with the `Learner[T]` interface.

src/main/scala/io/citrine/lolo/trees/classification/ClassificationTree.scala
```diff
@@ -121,10 +123,10 @@ class ClassificationTrainingResult(
   * Classification tree
   */
 class ClassificationTree(
-    rootModelNode: ModelNode[PredictionResult[Char]],
+    rootModelNode: ModelNode[Char],
     inputEncoders: Seq[Option[CategoricalEncoder[Any]]],
     outputEncoder: CategoricalEncoder[Any]
-) extends Model[ClassificationResult] {
+) extends Model[Any] {
```
sfriedowitz, line 129: This is one of the confusing parts I mentioned in the top level comments. A `ClassificationTree` is a `Learner[Any]` because it can train on any type of label data for classification. But the node structures that actually defines it is a bunch of `ModelNode[Char]` types because we encode the labels with a `CategoricalEncoder[Any]`. 

It works, but I can see why this is confusing.

bfolie:
> I'm not super familiar with the python release process. Do we need a version bump on that now, or only when we cut a Lolo release?

Bump the lolopy version as part of this PR. Then when you cut the release on GitHub, it sees that the version has been bumped and pushes to PyPi

python/lolopy/learners.py
```diff
@@ -351,38 +354,142 @@ def __init__(self, num_trees=-1, use_jackknife=True, bias_learner=None,
...
+            leaf_learner,
+            self.subset_strategy,
+            self.min_leaf_instances,
+            self.max_depth,
+            self.randomize_pivot_location,
+            self.randomly_rotate_features
+        )
+
+
+class MultiTaskRandomForest(BaseLoloRegressor):
```
bfolie, line 434: I don't love that there are now more options for the LoloPy user to choose from

src/main/scala/io/citrine/lolo/Model.scala
```diff
@@ -2,15 +2,15 @@ package io.citrine.lolo
 
 import breeze.linalg.DenseMatrix
 
-trait Model[+T <: PredictionResult[Any]] extends Serializable {
+trait Model[+T] extends Serializable {
```
bfolie, line 5: This is a good change

src/main/scala/io/citrine/lolo/trees/classification/ClassificationTree.scala
bfolie: It makes sense to me, but I've spent a lot of time in this code lately

sfriedowitz: I may encode this behavior a bit more firmly in one of the follow-up PRs. I've been thinking of making `Learner[T]` sealed and expliictly defining `RegressionLearner`, `ClassificationLearner`, etc..

I could imagine a scenario in which it makes sense to define `ClassificationLearner <: Learner[Any]` and `LeafClassifier <: Learner[Char]` as variants of the base trait.

python/lolopy/learners.py
sfriedowitz: I don't think there is an alternative option. The `_make_learner` method is called before the training data is encountered, so we would not know whether to construct a single- or multi-task RF on the scala side.

And I would argue that having them split on the Scala side is appropriate. Previously, the single- and multi-task RFs were both jammed in as a `RandomForest <: Learner`, which was distinctly not a `MultiTasklearner` within the Scala inheritance chain. I think breaking them up follows a much more intuitive inheritance chain for the multi-task option.

src/main/scala/io/citrine/lolo/Model.scala
sfriedowitz: Agreed. The generic now always means "the type of label data" across the package. I don't know why this ever deviated from that.

src/test/scala/io/citrine/lolo/DataGenerator.scala
```diff
@@ -0,0 +1,113 @@
...
+
+  /** Sequence of inputs and labels generated for training a learner. */
+  case class TrainingData[I, L](data: Seq[(Vector[I], L)])
```
bfolie, line 11: Putting a type parameter on the input doesn't seem like it gains us anything

bfolie:
Other than the one comment about changing `TrianingData[I, L]` to `TrainingData[L]`, this LGTM

src/test/scala/io/citrine/lolo/DataGenerator.scala
sfriedowitz: There is one test that relied on the fact that the initial inputs are a `Vector[Double]`. But I'm fine just casting for that one test for the sake of a cleaner trait.

sfriedowitz: I lied. There's are actually a few tests that rely on the fact that `DataGenerator.generate` initially produces continuous data of type `Seq[(Vector[Double], Double)]`. If we remove the type parameter `I` on the inputs, then we need to store in the training data class a `Seq[(Vector[Any], L)]` and resort to casting for when the inputs are required to be `Double`...which partially defeats the purpose of this interface.

StevenKauwe: @sfriedowitz, that makes sense to me. Is this something that you would come back to at a later date (because the additional code changes are out of scope), or is this just the best way for writing the code? Maybe @bfolie has some additional thoughts on this?

bfolie: Why do the tests require `Vector[Double]`? Are there any instances in the main code for which we want to restrict the input type? Making the interface more complex _solely_ to satisfy a test makes me suspicious.

sfriedowitz: > Are there any instances in the main code for which we want to restrict the input type

This interface is solely used as a data generator for test code. It never appears in the package source.

> Why do the tests require Vector[Double]?

A variety of reasons. A couple of tests jitter the data surrounding an input feature by adding `+/-` some noise. Another feeds an input feature into one of the Friedman functions. And many tests simply "bin" the input features, which requires them to be of type double.

sfriedowitz: The test code is littered with the following steps:
- Generate continuous input data of type Seq[(Vector[Double], Double)]`
- Transform either the input features or label values in some way
- Train a model on the transformed data, which must conform to the learner input type of `Seq[(Vector[Any], T)]`.

A generic on the input data is to allow for transformations in step 2 that operate on continuous (type `Double)` input data. Not because the `Learner[T]` interface is unable to handle data with `Vector[Any]` inputs.

bfolie: > This interface is solely used as a data generator for test code. It never appears in the package source.

Alright. As long as it's confined to the test code and doesn't become the interface for training data rows (for which I assume we'll want to have a label type parameter) then I'm OK with this

bfolie (approved):

PR #295, opened 2022-10-17
Concrete implementations for regression and classification

sfriedowitz:
This breaks apart a few generic learners that were not actually implemented in a generic manner (i.e., they required matching and casting internally to perform different operations on real/categorical data). Here is a list of changes:

1) Implements a `Standardizer[T]` interface with concrete implementations for `RegressionStandardizer <: Standardizer[Double]` and `ClassificationStandardizer[T] <: Standardizer[T]`.
2) Implements a `Bagger[T]` interface with concrete implementations for `RegressionBagger <: Bagger[Double]` and `ClassificationBagger[T] <: Bagger[T]`.
3) Moves the `MultiTaskETC...` learners into their own files in the standardizer/rotator/bagger modules
4) Misc updates to methods for readability, variable renames for consistency.

There are at least two more TODOs into this feature branch:
1) Scala naming conventions for accessor syntax on the core API
2) Implement a `TrainingRow[T]` class for how data is passed into the learners.


src/main/scala/io/citrine/lolo/bags/Bagger.scala
```diff
@@ -100,191 +115,102 @@ case class Bagger[T: ClassTag](
...
+  * @param disableBootstrap whether to disable bootstrap (useful when `method` implements its own randomization)
+  * @tparam T the type of label data
   */
-class BaggedModel[+T: ClassTag](
-    models: ParSeq[Model[T]],
-    Nib: Vector[Vector[Int]],
-    useJackknife: Boolean,
-    biasModel: Option[Model[Double]] = None,
-    rescale: Double = 1.0,
+case class ClassificationBagger[T](
```
sfriedowitz, line 180: I have been debating whether to make this class `ClassificationBager[T] <: Bagger[T]` or `ClassificationBagger <: Bagger[Any]`. 

My initial thought was that since this class is wrapping a `baseLearner[T]`, we should respect and carry through that type of label. But in practice, this is only used to wrap a `ClassificationTree <: Learner[Any]`, so the generic is not strictly necessary to make the code compile.

sfriedowitz: The same consideration applies to my implementation of `ClassificationStandardizer[T] <: Standardizer[T]`.

src/main/scala/io/citrine/lolo/bags/BaggerHelper.scala
```diff
@@ -1,71 +1,56 @@
...
   * @param models collection of trained models
   * @param trainingData on which models were trained
   * @param Nib vector (over models) of vectors (over training data) of the number of repeats in each model's bag
   * @param useJackknife whether to use jackknife for uncertainty quantification
   * @param uncertaintyCalibration whether to apply empirical uncertainty calibration
   */
-protected case class BaggerHelper(
-    models: ParSeq[Model[Any]],
-    trainingData: Seq[(Vector[Any], Any)],
+protected[bags] case class BaggerHelper(
```
sfriedowitz, line 17: The only active path through this helper class was on regression tasks, where it was used to calculate OOB errors and the uncertainty rescale. After breaking up the regression and classification baggers, I was able to make this helper type-specific for real-valued data.

src/main/scala/io/citrine/lolo/bags/MultiTaskBagger.scala
```diff
@@ -131,25 +138,23 @@ case class MultiTaskBagger(
...
+  * @param trainingData      multi-label training data
+  * @param biasModels        sequence of optional bias-correction models, one for each label
+  * @param rescaleRatios     sequence of uncertainty calibration ratios for each label
   */
 class MultiTaskBaggedTrainingResult(
     models: ParSeq[MultiTaskModel],
-    featureImportance: Option[Vector[Double]],
     Nib: Vector[Vector[Int]],
     trainingData: Seq[(Vector[Any], Vector[Any])],
-    useJackknife: Boolean,
```
sfriedowitz, line 147: Parameter dropped because it was unused in the subsequent code.

src/main/scala/io/citrine/lolo/transformers/standardizer/Standardizer.scala
```diff
@@ -0,0 +1,64 @@
+package io.citrine.lolo.transformers.standardizer
+
+import io.citrine.lolo.Learner
+import io.citrine.random.Random
+
+trait Standardizer[T] extends Learner[T] {
```
sfriedowitz, line 6: The `Standardizer[T]` and associated interfaces don't offer much in terms of code; I didn't see much between regression and classification that could be generically moved to the interface. But I still opted to implement a trait just to unify the inheritance chain between regression and classification standardizers.

src/main/scala/io/citrine/lolo/transformers/standardizer/StandardizerPrediction.scala
```diff
@@ -0,0 +1,64 @@
...
+  override def getExpected(): Seq[Double] = basePrediction.getExpected().map(outputTrans.invert)
+
+  // TODO: A PredictionResult[Double] should always return a Option[Seq[Double]] for uncertainty
+  override def getUncertainty(includeNoise: Boolean = true): Option[Seq[Any]] = {
+    basePrediction.getUncertainty(includeNoise).map { x =>
+      x.map(_.asInstanceOf[Double] * outputTrans.scale)
+    }
+  }
+
+  override def getGradient(): Option[Seq[Vector[Double]]] = {
```
sfriedowitz, line 26: Unfortunately these are in new files and not picked up by the git diff. I did make some changes in how the `Option` is handled when implementing a handful of these methods, but the logic is intended to remain the same.

I think the `Standardizer` tests passing unchanged is a good hint that we didn't make a mistake here.

src/main/scala/io/citrine/lolo/transformers/standardizer/Standardization.scala
```diff
@@ -0,0 +1,108 @@
...
+/** A package of the shift/scale required to standardize values to zero mean and unit variance. */
+case class Standardization(shift: Double, scale: Double) {
+  require(scale > 0 && scale < Double.PositiveInfinity)
+
+  def apply(x: Double): Double = (x - shift) / scale
+
+  def invert(x: Double): Double = x * scale + shift
+}
+
+object Standardization {
```
sfriedowitz, line 12: Same helper functions as what used to be on the `Standardizer` companion object, but renamed.

src/main/scala/io/citrine/lolo/bags/BaggedModel.scala
```diff
@@ -0,0 +1,76 @@
...
+        Some(a.get + b.get)
+      }
+
+      val scale = 1.0 / ensembleShapley.length
+      Some(scale * ensembleShapley.reduce(sumReducer).get)
+    }
+  }
+}
+
+class BaggedRegressionModel(
```
bfolie, line 33: Can these be case classes?

sfriedowitz: Yeah, it's pretty inconsistent atm. I'll make them all case class.

src/test/scala/io/citrine/lolo/transformers/rotator/MultiTaskFeatureRotatorTest.scala
```diff
@@ -0,0 +1,98 @@
...
+import io.citrine.lolo.{DataGenerator, SeedRandomMixIn}
+import org.junit.Test
+
+class MultiTaskFeatureRotatorTest extends SeedRandomMixIn {
+
+  /**
+    * Verify that rotated-input multi-task trees have expected properties
+    */
+  @Test
+  def testMultiTaskRotator(): Unit = {
```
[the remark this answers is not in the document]
sfriedowitz: I did not intend to make any edits to tests here besides what was necessary to make them compile. So in this case yes, it should be the same as before.

src/main/scala/io/citrine/lolo/bags/Bagger.scala
bfolie: By making it extend `Bagger[T]` you allow for the possibility of someone instantiating, say, `ClassificationBagger[Boolean]` or some other more specific label type. But we don't see any reason to do that. Are there any other practical distinctions between the two choices?

If it's such a small difference, then I don't have a strong opinion either way. I suppose I lean towards extending `Bagger[Any]` based on the principle of "don't create more options than you need." But it's a weak preference.

bfolie:
I like seeing so much `asInstanceOf` and `isInstanceOf` being removed. I agree with the way you've broken up the files and combined shared code. Given our desire to push towards more strongly typed learners I think that this is the clear way to do it.

I just have one outstanding question about `ClassificationBagger` extending `Learner[T]` vs. `Learner[Any]`.

src/main/scala/io/citrine/lolo/bags/Bagger.scala
sfriedowitz: The thought I'm leaning towards as of now is that since this bagger class wraps a `Learner[T]` we should respect the base learner's type, and produce a `TrainingResult[T]`, `Model[T]`, and `PredictionResult[T]`, rather than "erasing" the type into a `Learner[Any]` on instantiation. I also that since `Learner[T]` is invariant in its type parameter, that "erasing" to `Any` would actually not work, although I can test this..

Either way, since the code is actually implemented generically now for `ClassificationBagger[T]`, I still would opt to keep it unless you object otherwise.

bfolie (approved):

src/main/scala/io/citrine/lolo/bags/Bagger.scala
StevenKauwe: @sfriedowitz, regarding this. Didn't you say in practice the classification data always comes in as a string?

sfriedowitz: It does from Mithril, at least, where the label type is derived from the associated descriptor value types. This library is a bit more generic.

sfriedowitz: One of the considerations here is that within the `trees` code, categorical is encoded with a function that maps `T => Char`, and the actual classification learning is performed on the low-memory `Char` data type. Some of these `Learner` implementations are required to operate on either top-level `T`-valued data, or low-level `Char`-valued data. So in those cases, the generics are necessary (e.g., `GuessTheModeLearner`).

PR #296, opened 2022-10-19
PLA-10433: Training row class

sfriedowitz:
Adds the `TrainingRow[T]` class as input for a `Learner[T]` training, and cascades updates throughout the library to pass around these structures types instead of tuples of (inputs, label, weight).

### DIFF
--- a/src/main/scala/io/citrine/lolo/TrainingRow.scala
+++ b/src/main/scala/io/citrine/lolo/TrainingRow.scala
@@ -1,11 +1,24 @@
 package io.citrine.lolo
 
-case class TrainingRow[+T](inputs: Vector[Any], labels: T, weight: Double) {
+case class TrainingRow[+T](inputs: Vector[Any], label: T, weight: Double) {
+
+  def asTuple: (Vector[Any], T, Double) = (inputs, label, weight)
+
+  def withInputs(inputs: Vector[Any]): TrainingRow[T] = copy(inputs = inputs)
+
   def withWeight(weight: Double): TrainingRow[T] = copy(weight = weight)
 }
 
 object TrainingRow {
 
+  /**
+    * Extract the `idx` label from a multi-task training row and cast it to type `T`.
+    *
+    * @param row containing multi-task label vectors
+    * @param idx of the label to extract
+    * @tparam T of the extracted label
+    * @return a TrainingRow[T] for the extracted label type
+    */
   def extractLabel[T](row: TrainingRow[Vector[Any]], idx: Int): TrainingRow[T] =
-    TrainingRow(row.inputs, row.labels(idx).asInstanceOf[T], row.weight)
+    TrainingRow(row.inputs, row.label(idx).asInstanceOf[T], row.weight)
 }
--- a/src/main/scala/io/citrine/lolo/TrainingRow.scala
+++ b/src/main/scala/io/citrine/lolo/TrainingRow.scala
@@ -1,15 +1,43 @@
 package io.citrine.lolo
 
-case class TrainingRow[+T](inputs: Vector[Any], label: T, weight: Double) {
+case class TrainingRow[+T](inputs: Vector[Any], label: T, weight: Double = 1.0) {
 
   /** Cast the training row as a tuple of (inputs, label, weight). */
   def asTuple: (Vector[Any], T, Double) = (inputs, label, weight)
+
+  /** Return a new row updated with the provided inputs. */
+  def withInputs(inputs: Vector[Any]): TrainingRow[T] = copy(inputs = inputs)
+
+  /** Return a new row updated with the provided label. */
+  def withLabel[T2](label: T2): TrainingRow[T2] = copy(label = label)
+
+  /** Return a new row updated with the provided weight. */
+  def withWeight(weight: Double): TrainingRow[T] = copy(weight = weight)
+
+  /** Return a new row with the function applied to the original inputs. */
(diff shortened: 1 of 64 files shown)


src/main/scala/io/citrine/lolo/trees/classification/ClassificationTrainingLeaf.scala
```diff
@@ -1,19 +1,19 @@
...
-import io.citrine.lolo.{Learner, TrainingResult}
+import io.citrine.lolo.{Learner, TrainingResult, TrainingRow}
 import io.citrine.lolo.trees.TrainingLeaf
 import io.citrine.random.Random
 
 import scala.collection.mutable
 
 /** Training leaf for a classification tree. */
 case class ClassificationTrainingLeaf(
-    trainingData: Seq[(Vector[AnyVal], Char, Double)],
```
sfriedowitz, line 11: One thing to note here. Because `TrainingRow` contains `inputs: Vector[Any]`, we are sacrificing the typing within trees that the inputs are a `Vector[AnyVal]` after encoding. 

Seeing as the distinction between `Any` and `AnyVal` doesn't really offer us much here, I think this is a fair trade to use the same `TrainingRow` class in all locations.

src/main/scala/io/citrine/lolo/TrainingRow.scala
```diff
@@ -0,0 +1,41 @@
...
+  def mapLabel[T2](f: T => T2): TrainingRow[T2] = copy(label = f(label))
+
+  /** Return a new row holding up to the first `n` features from the input vector. */
+  def takeInputs(n: Int): TrainingRow[T] = copy(inputs = inputs.take(n))
+}
+
+object TrainingRow {
+
+  /** Build a sequence of training from the provided data and weights. */
+  def build[T](data: Seq[(Vector[Any], T)], weights: Option[Seq[Double]] = None): Seq[TrainingRow[T]] = {
```
sfriedowitz, line 38: I wanted to make some `implicit class`es that map `(Vector[Any], T) => TrainingRow[T]` and so forth. This would allow for a syntax like `inputs.zip(labels).map(_.asTrainingRow)`.

But, the compiler was having trouble importing the implicits defined within the `TrainingRow` companion object. So I gave up and am just using the helper method explicitly.

src/test/scala/io/citrine/lolo/DataGenerator.scala
```diff
@@ -8,33 +8,32 @@ import io.citrine.random.Random
...
-      val (baseInputs, baseLabels) = trainingData.data.unzip
+  case class TrainingData[T](data: Seq[TrainingRow[T]]) {
+    def withBinnedInputs(bins: Seq[(Int, Int)]): TrainingData[T] = {
+      val baseInputs = data.map(_.inputs)
       val indexToBins = bins.toMap
       val transposedInputs = baseInputs.transpose
       val binnedInputs = Vector
         .tabulate(transposedInputs.length) { index =>
-          val indexData = transposedInputs(index)
+          val indexData = transposedInputs(index).asInstanceOf[Seq[Double]]
```
sfriedowitz, line 18: We have to cast here now when binning data for various unit tests, but its exclusively used within the test code so :shrug:

src/main/scala/io/citrine/lolo/trees/classification/ClassificationTrainingLeaf.scala
bfolie: Why can't we enforce `inputs: Vector[AnyVal]` throughout? Do we allow an input entry to be a sequence?

sfriedowitz: A `String` is not an `AnyVal`, is the main reason.

src/main/scala/io/citrine/lolo/bags/BaggerHelper.scala
```diff
@@ -28,21 +28,21 @@ protected[bags] case class BaggerHelper(
     */
   lazy val oobErrors: Seq[(Vector[Any], Double, Double)] = trainingData.indices.flatMap { idx =>
     val oobModels = models.zip(Nib.map(_(idx))).filter(_._2 == 0).map(_._1)
-    val label = trainingData(idx)._2
+    val label = trainingData(idx).label
```
bfolie, line 31: I love seeing `._1` and `._2` replaced with `.inputs` and `.label`

src/main/scala/io/citrine/lolo/trees/classification/ClassificationTree.scala
```diff
@@ -27,52 +27,43 @@ case class ClassificationTreeLearner(
...
     /* Add the weights to the (features, label) tuples and remove any with zero weight */
-    val finalTraining = encodedTraining
-      .zip(weights.getOrElse(Seq.fill(trainingData.size)(1.0)))
-      .map {
-        case ((f, l), w) =>
-          (f, l, w)
-      }
-      .filter(_._3 > 0)
-      .toVector
+    val finalTraining = encodedTraining.filter(_.weight > 0.0)
```
bfolie, line 60: :rocket:

src/test/scala/io/citrine/lolo/DataGenerator.scala
bfolie: You could also create a `RealValuedTrainingData` class that enforced the inputs as `Double`. But then it would have to have some function to cast to the `TrainingRow` type and you'd be calling that function dozens of times across many tests.
I think that this single `asInstanceOf` is a more reasonable price to pay.

bfolie (approved):

PR #298, opened 2022-10-25
Move API to api folder

sfriedowitz:
A straight refactor, moving the top-level interfaces into their own `api` folder. The only code update is in `learners.py` to correct a path calling `TrainingRow.build`.

### DIFF
--- a/python/lolopy/learners.py
+++ b/python/lolopy/learners.py
@@ -92,7 +92,7 @@ class BaseLoloLearner(BaseEstimator, metaclass=ABCMeta):
         assert weights_java.length() == len(X), "Weights copy failed"
 
         # Train the model
-        train_rows = self.gateway.jvm.io.citrine.lolo.TrainingRow.build(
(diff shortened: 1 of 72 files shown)


sfriedowitz:
Failed due to some dependency download issue...rebuilding

StevenKauwe:
@sfriedowitz, looks like the rebuild succeeded this time. :D

sfriedowitz:
@StevenKauwe it did indeed...so this seems ready to merge pending your approval 😉

StevenKauwe (approved):
LGTM.
quick question. where do these live?
`io.citrine.lolo.{DataGenerator, SeedRandomMixIn, etc.}`

sfriedowitz:
>quick question. where do these live?
>io.citrine.lolo.{DataGenerator, SeedRandomMixIn, etc.}

These are in the test code. So instead of `src/main`, they are in `src/test`, but the package path is mimicked on both sides.

PR #301, opened 2022-11-01
PLA-10549: Multi-task linear regression

sfriedowitz:
Adds a multi-task linear regression learner that trains independent single-task learners on each label.

### DIFF
--- a/src/main/scala/io/citrine/lolo/linear/MultiTaskLinearRegression.scala
+++ b/src/main/scala/io/citrine/lolo/linear/MultiTaskLinearRegression.scala
@@ -6,18 +6,25 @@ import io.citrine.random.Random
 case class MultiTaskLinearRegressionLearner(regParam: Option[Double] = None, fitIntercept: Boolean = true)
     extends MultiTaskLearner {
 
-  override def train(trainingData: Seq[TrainingRow[Vector[Any]]], rng: Random): MultiTaskTrainingResult = {
+  override def train(
+      trainingData: Seq[TrainingRow[Vector[Any]]],
+      rng: Random
+  ): MultiTaskLinearRegressionTrainingResult = {
     val rep = trainingData.head
     val repLabels = rep.label
     val numLabels = repLabels.length
-    assert(repLabels.forall(_.isInstanceOf[Double]))
+    assert(
+      repLabels.forall(_.isInstanceOf[Double]),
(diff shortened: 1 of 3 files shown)


src/main/scala/io/citrine/lolo/linear/MultiTaskLinearRegression.scala
```diff
@@ -0,0 +1,49 @@
+package io.citrine.lolo.linear
+
+import io.citrine.lolo.api.{MultiTaskLearner, MultiTaskModel, MultiTaskTrainingResult, ParallelModels, TrainingRow}
+import io.citrine.random.Random
+
+case class MultiTaskLinearRegressionLearner(regParam: Option[Double] = None, fitIntercept: Boolean = true)
```
sfriedowitz, line 6: I could definitely imagine cases in which someone would want to specify a single regParam per label. But our `Learner`s (and the associated settings structures in Mithril) are defined such that they do not need to know the # of labels when constructed. So that would require some slightly annoying validation on # params = # labels.

So I just left it at a single parameter for now.

src/main/scala/io/citrine/lolo/linear/MultiTaskLinearRegression.scala
```diff
@@ -0,0 +1,49 @@
...
+
+  override def train(
+      trainingData: Seq[TrainingRow[Vector[Any]]],
+      rng: Random
+  ): MultiTaskLinearRegressionTrainingResult = {
+    val rep = trainingData.head
+    val repLabels = rep.label
+    val numLabels = repLabels.length
+    assert(
+      repLabels.forall(_.isInstanceOf[Double]),
```
sfriedowitz, line 17: We still have a bit of a failure in the `MultiTaskLearner` interface such that it expects a learner to act on a label `Vector[Any]`, but this one is only defined for multi-task regression.

I don't want to go through that refactor right now so just going to define as this.

bfolie (approved):

src/main/scala/io/citrine/lolo/linear/MultiTaskLinearRegression.scala
bfolie: In the long run in Mithril it might make more sense for us to spin up a bunch of single-task linear learners instead of using multi-task linear learner, because then we can optimize the regularization parameter of each one individually. But this will work for now

PR #307, opened 2023-03-22
Create error types for not enough data

bfolie:
Having structured error messages here will help us catch the issue and return more actionable messages to users

### DIFF
--- a/src/main/scala/io/citrine/lolo/bags/BaggedTrainingError.scala
+++ b/src/main/scala/io/citrine/lolo/bags/BaggedTrainingError.scala
@@ -9,14 +9,14 @@ case class InsufficientTrainingDataException(
     numRows: Int,
     numRequired: Int
 ) extends BaggedTrainingError(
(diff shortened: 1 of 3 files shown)


src/main/scala/io/citrine/lolo/bags/BaggedTrainingError.scala
```diff
@@ -0,0 +1,22 @@
+package io.citrine.lolo.bags
+
+sealed trait BaggedTrainingError {
```
sfriedowitz, line 3: I think that this trait should extend `Exception` directly.

You can follow the usual pattern of making this a `sealed abstract class BaggedTrainingError(msg: String)` such that the message can be passed along to the base exception class.

bfolie: I realized something similar while testing locally, although I made it work by extending Throwable. Switching to sealed abstract class now...

sfriedowitz (approved):
GLTM

PR #308, opened 2023-03-23
bump sbt-ci-release version

bfolie:


sfriedowitz (approved):
Not sure the implications here, but LGTM

bfolie:
> Not sure the implications here, but LGTM

Clearly I wasn't either because it took me 5 tries to get a 2 line diff correct 😬 

The release to Maven Central is failing. I generated new keys and verified the Sonatype credentials, so the only other thing I can think of is to update the sbt-ci-release plugin.

PR #309, opened 2023-03-29
Bump Scala to 2.13.10

bfolie:


sfriedowitz (approved):
We can probably delete crossScalaVersions since we arent cross compiling anymore.

bfolie:
> We can probably delete crossScalaVersions since we arent cross compiling anymore.

I agree. But I want to make only one change at a time until we get this build issue fixed

PR #318, opened 2024-09-24
Add save and load method

mVenetos97:
#### Ticket: PNE-6225

Lolopy models do not currently have any methods to serialize/deserialize making them difficult to work with. This change adds save/load methods.


StevenKauwe:
Can you add a test for the save and load methods you have added?

StevenKauwe (approved):
lgtm

PR #320, opened 2024-09-24
PNE-6228: Replace Travis CI with Github Actions

mVenetos97:
Move from travis CI to github actions


.github/workflows/merge_workflow.yml
```diff
@@ -0,0 +1,103 @@
...
+
+env:
+  SCALA_VERSION: 2.13.8
+  JAVA_VERSION: 11
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
```
StevenKauwe, line 18: Using a matrix here feels a bit awkward since you need custom logic for each. Might be more clear to have two separate jobs that run in parallel.

.github/workflows/merge_workflow.yml
```diff
@@ -0,0 +1,103 @@
...
+        test-type: ['scala', 'python']
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up JDK
+      if: matrix.test-type == 'scala'
+      uses: actions/setup-java@v2
+      with:
+        java-version: ${{ env.JAVA_VERSION }}
```
StevenKauwe, line 26: are these env vars set? I don't see them. Same with the secrets you reference.

.github/workflows/merge_workflow.yml
```diff
@@ -0,0 +1,103 @@
...
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up JDK
+      uses: actions/setup-java@v2
+      with:
+        java-version: ${{ env.JAVA_VERSION }}
+        distribution: 'adopt'
+    - name: Release to Maven Central
+      env:
+        PGP_PASSPHRASE: ${{ secrets.PGP_PASSPHRASE }}
```
StevenKauwe, line 79: what are these secrets? Have you set them? How do they get used?

.github/workflows/merge_workflow.yml
```diff
@@ -0,0 +1,103 @@
...
+    - name: Install dependencies for PyPI release
+      run: |
+        sudo apt-get install -y pandoc
+        cd python
+        make
+        pip install --only-binary=numpy,scipy -r requirements.txt
+    - name: Build and publish to PyPI
+      env:
+        TWINE_USERNAME: __token__
+        TWINE_PASSWORD: ${{ secrets.PYPI_API_TOKEN }}
```
StevenKauwe, line 97: Is this the same token as from the repo used by our other python libraries (citrine-python & gemd-python)?

.github/workflows/tests.yml
```diff
@@ -39,8 +39,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
-        os: [ubuntu-latest, windows-latest, macos-latest]
+        python-version: ["3.8"]
+        os: [ubuntu-latest]
```
mVenetos97, line 43: An earlier commit had python 3.8-3.12 and also had mac and windows OS. Unfortunately, it looks like lolo is not up to date so I reverted back to ubuntu and 3.8. I plan to make another PR to bring lolo into the present

.github/workflows/merge_workflow.yml
mVenetos97: Good point, I split into separate jobs.

.github/workflows/merge_workflow.yml
mVenetos97: No, I just have them set in the yml itself

.github/workflows/merge_workflow.yml
mVenetos97: These are for deploying and they are set in Github

.github/workflows/merge_workflow.yml
mVenetos97: Yes but I dont think you can share them across our libraries

StevenKauwe (approved):
lgtm?

PR #321, opened 2024-09-30
Pne 6240 optimize shapley

mVenetos97:
Optimizing Shapley calculations to run faster. Two main changes:

`BaggedModel.scala`:
The `.shapley.` method in `BaggedModel` is calling `.shapley` on each model in the ensemble and then summing the results. This approach is optimized by:
- parallelizing (`par`) Shapley calculation for each model.
- Simplify the reduction step by directly adding matrices instead of using a separate reducer function.
- Avoid creating intermediate `Option` objects during reduction.

`ModelNode.scala`
Small optimizations:
- Precompute `hotPortion` and `coldPortion` in InternalModelNode to avoid repeated calculation.

These two changes brought the time from ~30s down to ~7s for a dataset of 100 points. (tested locally)


src/main/scala/io/citrine/lolo/bags/BaggedModel.scala
```diff
@@ -14,18 +14,19 @@ trait BaggedModel[+T] extends Model[T] {
   override def transform(inputs: Seq[Vector[Any]]): BaggedPrediction[T]
 
   override def shapley(input: Vector[Any], omitFeatures: Set[Int] = Set()): Option[DenseMatrix[Double]] = {
-    val ensembleShapley = ensembleModels.map(model => model.shapley(input, omitFeatures))
-    if (ensembleShapley.head.isEmpty) {
+    val ensembleShapley: Seq[Option[DenseMatrix[Double]]] = ensembleModels.par.map(_.shapley(input, omitFeatures)).seq
```
mVenetos97, line 17: parallelizing (par) Shapley calculation for each model.

src/main/scala/io/citrine/lolo/bags/BaggedModel.scala
```diff
@@ -14,18 +14,19 @@ trait BaggedModel[+T] extends Model[T] {
...
     } else {
-      assert(ensembleShapley.forall(x => x.isDefined))
-
-      def sumReducer(a: Option[DenseMatrix[Double]], b: Option[DenseMatrix[Double]]): Option[DenseMatrix[Double]] = {
-        Some(a.get + b.get)
+      val validMatrices: Seq[DenseMatrix[Double]] = ensembleShapley.flatten
+      if (validMatrices.isEmpty) {
+        None
+      } else {
+        val sumMatrix: DenseMatrix[Double] = validMatrices.reduce(_ + _)
```
mVenetos97, line 26: Simplify the reduction step by directly adding matrices instead of using a separate reducer function.

src/main/scala/io/citrine/lolo/bags/BaggedModel.scala
```diff
@@ -14,18 +14,19 @@ trait BaggedModel[+T] extends Model[T] {
...
+        val scale = 1.0 / ensembleModels.length
+        Some(scale * sumMatrix)
```
mVenetos97, line 28: Avoid creating intermediate Option objects during reduction.

src/main/scala/io/citrine/lolo/trees/ModelNode.scala
```diff
@@ -101,27 +104,16 @@ case class InternalModelNode[+T](
...
-    } else {
-      (right, left)
-    }
-
     // If the feature is omitted or unknown, then the weight assigned to each child is taken
     // as the share of the training data that got split into each child
-    val hotPortion = hot.trainingWeight / trainingWeight
-    val coldPortion = cold.trainingWeight / trainingWeight
+    val (hot, cold) = if (goLeft) (left, right) else (right, left)
+    val (hotPortion, coldPortion) = if (goLeft) (leftPortion, rightPortion) else (rightPortion, leftPortion)
```
mVenetos97, line 112: condensing the previous code which figured out the hot and cold paths then computed the Portions. Now we compute portions once and assign when necessary.

anoto-moniz (approved):
This looks great to me!

PR #325, opened 2024-10-28
Pne 6240 add transient decorator to shapley

mVenetos97:
This change adds a `@transient lazy private val` decorator to the `leftPortion` and `rightPortion` in `ModelNode`. This keeps us from serializing the portions which will break SerDe


StevenKauwe (approved):
lgtm

PR #328, opened 2025-02-07
pne-6628/fix-pypi-deploy

mVenetos97:
Adding an sbt installation so we can build lolo so that lolopy has the necessary jar files


kroenlein (approved):
LGTM

PR #330, opened 2025-02-10
Update mixin order per scikit-learn 1.6 release

kroenlein:
With the Tag abstraction introduced in scikit-learn 1.6, method resolution order grabs the tags from [BaseEstimator](https://github.com/scikit-learn/scikit-learn/blob/d8932866b6f4b2dee508a54b79f1122ff5f5459d/sklearn/base.py#L421) instead of [RegressorMixin](https://github.com/scikit-learn/scikit-learn/blob/d8932866b6f4b2dee508a54b79f1122ff5f5459d/sklearn/base.py#L614), resulting in a py4j casting error on targets.


mVenetos97 (approved):
LGTM

PR #332, opened 2025-02-13
Update testing to use os / version matrix

kroenlein:
Updates the actions testing to include a range of python versions and to test 3.12 on a range of os's.  Also updates requirements to match the actual range of Python versions.


mVenetos97 (approved):
LGTM

PR #333, opened 2025-02-13
Modify publish actions so publishing happens upon releases

kroenlein:
This modifies both publish actions so that they only run when we manually trigger a release.

This also homogenizes the various actions so that dependencies between the various actions are consistent.


mVenetos97 (approved):
Lgtm

PR #334, opened 2025-02-13
Drop 3.8 support to resole dependabot alert

kroenlein:
https://github.com/CitrineInformatics/lolo/security/dependabot/1

Support for 3.8 required potentially depending on a vulnerable version of scikit-learn.


mVenetos97 (approved):
Looks good to me

PR #335, opened 2025-02-15
Rework Python tests to use pytest, enforce coverage

kroenlein:
This PR is primarily focused on improving tests on lolopy.  It should include no changes to functionality.

Note that coverage is skipped on recursive calls to `_make_learners` when there are user-defined learners present.  It was not clear how to exercise those routes.

* Migrate tests to use pytest, allowing for monkeypatching, and numpy.testing where appropriate
* Rename version.py to __version__.py, per standard as not a user-facing module
* Combine deploy actions into one script
* Update testing to require 100 % coverage for Python files
* Change `__get_state__` in `BaseLoloLearner` to always copy the state dictionary, as the result returned from latest sklearn uses a tied dictionary, so deleting gateway corrupted the object.  Also change the del to a pop so, removing the need for a try / except block.
* Add `__init__` to `BaseLoloClassifier` so that n_classes is not initialized outside of an `__init__` method
* Add testing for pickling and migrate testing for json serialization to use a temporary directory
* Migrate file handling from `os` to `pathlib`


mVenetos97 (approved):
LGTM

PR #336, opened 2025-02-17
Fix pickling for learners

### LINKED ISSUE #313: How to save the fitted RandomForestRegressor model?
How can I save a lolopy model?"

I tried to train a model like this:
```
from lolopy.learners import RandomForestRegressor
model = RandomForestRegressor()
model.fit(X, Y)
```
After that, I attempted to save the model using:
```
joblib.dump(model, "./model.pkl")
```
But, it didn't work with the following error:
```
AttributeError: 'RandomForestRegressor' object has no attribute 'gateway'
```

Thank you for sharing your great program.

WardLT:
Addresses two issues:

- Make a copy of the state dict so editing the state dict doesn't affect the underlying object
- A missing key is a "KeyError" not an "AttributeError"

Fixes #313

### DIFF
--- a/python/lolopy/learners.py
+++ b/python/lolopy/learners.py
@@ -48,14 +48,14 @@ class BaseLoloLearner(BaseEstimator, metaclass=ABCMeta):
     def __getstate__(self):
         # Get the current state
         try:
-            state = super(BaseLoloLearner, self).__getstate__()
+            state = super(BaseLoloLearner, self).__getstate__().copy()
         except AttributeError:
             state = self.__dict__.copy()
 
         # Delete the gateway data
         try:
             del state['gateway']
-        except AttributeError:
+        except KeyError:
             pass
 
         # If there is a model set, replace it with the JVM copy
--- a/python/lolopy/learners.py
+++ b/python/lolopy/learners.py
@@ -47,16 +47,10 @@ class BaseLoloLearner(BaseEstimator, metaclass=ABCMeta):
         
     def __getstate__(self):
         # Get the current state
-        try:
-            state = super(BaseLoloLearner, self).__getstate__().copy()
-        except AttributeError:
-            state = self.__dict__.copy()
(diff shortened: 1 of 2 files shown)


kroenlein:
@WardLT Hi Logan.  Looks like you and I noticed the same problem around the same time.  Buried inside #335, you might notice I made similar edits to python/lolopy/learners.py.  Do you know of the scenario that requires catching the AttributeError after invoking `super.__getstate__`?  It looks like the changes made to SKLearn that broke all this removed that risk.

And FYI, `pop` allows key removal from a dictionary without risking an exception, so swapping the del to a pop obviates the need for a try-catch.

WardLT:
Hello Ken!

> Do you know of the scenario that requires catching the AttributeError after invoking super.__getstate__? 

I've never seen this idiom before, and it seems like it's borrowed from [the sklearn `__getstate__` method's logic to avoid problems with multiple inheritance](https://github.com/scikit-learn/scikit-learn/pull/8324)

I think it'd be save to remove the try/catch block because we don't have any multiple inheritance in this part of lolopy.

> And FYI, pop allows key removal from a dictionary without risking an exception, so swapping the del to a pop obviates the need for a try-catch.

I'll go ahead and do that.

WardLT:
Also, I missed that you had already fixed this issue in #335! Hilarious timing that I happened to run into it at the same time.

Feel free to close my PR if it is indeed duplicative.

kroenlein:
I'll chalk it up to people finally waking up from their winter slumber and running `pip install -U`.  I'll be pushing a new release shortly, so PyPI will have pickling support again.  Let us know if you spot any other issues!

FYI: for the check version test that was failing, we have a script that verifies that the new version in lolopy/__version__.py is larger than the previous one.  A bump on patch is enough to satisfy it.

kroenlein:
Duplicate of #335

